PR #110, opened 2015-10-21
AES debugged and libutils updated

deekshadangwal:


Gamrix:
Could you replace your last commit to not remove the str_to_int_array function?

do on your branch:
> git reset --hard HEAD~

Then modify the libutils function without removing str_to_int_array and push it back to this branch

deekshadangwal:
Added str_to_int_array.

Gamrix:
Just for the record: The reason why this thread looks so weird is because there were modifications to existing commits (instead of a series of commits to fix things)

PR #112, opened 2015-10-26
Some changes to Example 1 to try to make = vs <<= clearer

Gamrix:
After seeing the struggle people had with differentiating <<= and =, I have reworked example 1 to try to make the differences clearer and easier to understand. However, as a person that is very familiar with PyRTL, I can sometimes be blinded by my experience. Therefore, I am asking all of you to read over the changes I have made and give me feedback on whether I was actually able to achieve my goal.


jemcmahan13:
The example might be too long now for a first example; it can be hard to follow. What's really good about the examples is that each has a really clear goal --- but with extra explanation one can lose track of what the point is.

<<= vs = is difficult to grasp for new users because it's actually a pretty subtle point, and different from both hardware and software languages. I'm not against elaborating on it, but the first example should be simple and clear. A statement circa line 54 in the original clarifying the difference would be good (and maybe sufficient). A small "review" section at the end to clarify complicated points could also be beneficial.

More specifically, 
-I think showing implicit bitwidths shouldn't be in the first example. The best practice is to always specify a bitwidth, and it's complicating the flow to demonstrate a minor --- but nice --- feature. Ditto for explaining working_block. 
-The copies and asserts feel a little aimless; along with the added explanation, the point gets a little lost.
-There are both lettered and enumerated tmp variables; it's not clear why.
-You made a second commit for a grammar correction but missed several other grammatical errors :)

PR #117, opened 2015-12-13
Add AES back into the development trunk

Gamrix:
AES looks like it is finally cleaned up to be merged back into the main trunk.

The code is much cleaner than what we started out with

Also, both the combinatorial and the state machine decryption blocks are functional. I do not think that encryption is needed before we merge back into trunk.

@timsherwood What do you think?


timsherwood:
I would like a chance to go through it first (hopefully later this week) just to make sure everything is looking sharp. :)

PR #128, opened 2016-02-06
Open-collector option for output-to-verilog

masoug:
Here is a shim I added to the verilog output function for PyRTL; it simply rewrites the original output wirevector as a normal verilog wire and creates a new verilog output that replaces HIGH signals (1) with high-impedance (z).

This is mainly useful for things like bidirectional outputs (such as required in I2C protocols) where one wire must be read from and written to. For more information, see here: https://en.wikipedia.org/wiki/Open_collector

Everything else in PyRTL should be the same, this does not affect PyRTL internals and is an implementation-specific parameter.

### DIFF
--- a/pyrtl/inputoutput.py
+++ b/pyrtl/inputoutput.py
@@ -453,13 +453,16 @@ def trace_to_html(simtrace, trace_list=None, sortkey=None):
 #    \/  |___ |  \ | |___ \__/ \__>
 #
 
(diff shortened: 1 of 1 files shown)


Gamrix:
As a single change to the Verilog Output, this does not seem generally useful enough to be added to the codebase. However, if there was a way to write plugins to the Verilog Output, I would be happy for this to be one such plugin in our plugin library.

Issue #118, opened 2015-12-17
Some Verilog-style constants not accepted in Const


jemcmahan13:
Const will accept Verilog style strings only if they specify no base or use binary or octal bases ('b' or 'o'). Bases 'h' and 'd' are all also valid in Verilog, but the function `_convert_verilog_str` relies on the Python built-in number base conversion, which supports 'b' and 'o', but not 'h' or 'd'.

```python
x = Const("5'b10011")  # accepted
x = Const("32'd12")  # rejected
x = Const("10'o7")  # accepted
x = Const("8'h7F")  # rejected
```

abe-k:
It looks like Verilog supports some other features in its integer literals as well (according to http://web.engr.oregonstate.edu/~traylor/ece474/lecture_verilog/beamer/verilog_number_literals.pdf). Do we want to support any of signedness, underscore-separators, or default values?

timsherwood:
It would be fine if signed is supported, but it should throw an error deeper in the const code (signed values is a discussion for another day).  It is also fine if we throw an error here for now.

Underscores are awesome and it would be great if they were supported

As to default widths I think we would be best to avoid supporting them -- explicit is probably better here and the idea that 32-bits is the default default is perhaps no longer valid (not sure if Verilog would make that same choice if they had to do it over).

abe-k:
I think I'll add support for negative literals (two's complement), but an error if the S prefix for signedness is used.

timsherwood:
Just be aware that there is some very unintuitive behavior if we add signed constants as twos complement without proper sign types:   e.g. Const('-1') > 0 is true; Const(-1).extended(32) will not be what you think; etc.

It probably merits some discussion while John is still around. :)

Perhaps it make sense to add a special "twos_complement" function or something like that in the mean time?  I agree that it is a PITB to have to do the two's complement calculation yourself though.

abe-k:
We already support negative integers being passed, as in `Const(-1, bitwidth=32)`. This just extends that support to Verilog-style strings, as in `Const("-32'd1")`.

timsherwood:
Sound good then --  thanks for clarifying.

timsherwood:
Is this one done?  Can we close this issue?

PR #133, opened 2016-04-20
Fix issue #118 with support for h, d, negatives, and underscores in string Consts

abe-k:
This makes the parsing of Verilog-style consts match more closely with actual Verilog behavior, and removes the dependence on the specifics of Python syntax as parsed by `int(x, 0)`.

### DIFF
--- a/pyrtl/wire.py
+++ b/pyrtl/wire.py
@@ -520,7 +520,7 @@ def _convert_verilog_str(val, bitwidth=None):
             sval = sval[1:]
         sval = sval.replace('_', '')
         num = int(sval, base)
-    except ValueError:
+    except (IndexError, ValueError):
         raise PyrtlError('error, string for Const not in verilog style format')
     if neg and num:
         if (num >> bitwidth-1):
--- a/pyrtl/wire.py
+++ b/pyrtl/wire.py
@@ -497,22 +497,27 @@ def _validate_const_int(val, bitwidth=None):
 
 
 def _convert_verilog_str(val, bitwidth=None):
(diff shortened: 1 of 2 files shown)


abe-k:
Looks like I have some tests to fix first. Do we want to drop support for `x` as an alias for `h`, or keep it?

Gamrix:
we would like to support the python way whenever possible, so please keep x

abe-k:
It's been a couple days—is there anything more I should do on this pull request before it can be accepted?

Gamrix:
Could you add more test cases to the unittest?
It would be great if all the new functionality is covered by tests. Some good tests would be:

- Test for error when you use an invalid base character
- Test for error when bitwidths are not long enough for a string with a base that is not 'b'
- Test for error when specifying a negative value with a base when no bitwidth is specified
- Test for error when bitwidths are both specified through the verilog string and the bitwidth parameter
- Test for proper parsing of values with underscores even when there is no bitwidth and/or no base

Also, could you get the following unittest working?

```python 
    @unittest.skip
    def test_badstring_broken(self):
        self.assert_bad_const("1'")
```

Gamrix:
@timsherwood I am wondering if we should actually support verilog 'negative' constant values.
They actually have a very weird format, eg. `-5'b10010` is `0b01110`

abe-k:
I think I've added all the tests you suggested. My handling of negative literals doesn't match the example you gave, though: it will raise an error if the constant is not positive before negation (high bit 0).

Gamrix:
Looks good! If @timsherwood doesn't have anything else to suggest, I'll pull it sometime tomorrow evening.

Issue #116, opened 2015-12-10
High level ops (eg '<', '*') not synthesizable


Gamrix:
I believe that sooner or later, we should implement synthesis for the multiplication op. One question would be which multiplication operator would we use by default @timsherwood .

Gamrix:
The main way to do this as I and @timsherwood have discussed is by having a presynthesis pass to reduce more complex hardware (like '+' and '*'). This shouldn't be hard once I move the transforms that are in PyRTL research into PyRTL core. Transforms will be very useful for this task and it will also be a good real world usage of that code. If anyone wants to take this on, please let me know, and I'll get code moved around to facilitate fixing this.

abe-k:
It looks like synthesis already handles `+`, `<`, etc., and that `*` is the only operator still unsynthesizable.

The `_basic_mult` provided uses `+`, so either it would have to be rewritten to only use simple single-bit operations, or it would have to be applied before `_decompose`.

Does anyone see a problem with (at the beginning of `synthesize`) copying the input block and using `net_transform` to replace the `*` operators with `_basic_mult`, and then using this modified block for the remaining synthesis?

Gamrix:
Nope, I don't see any problems with that.

Part of what we (ok, really I) want is to not have implementations of '+-<>=' and possibly 'x' that can only be used for synthesis. The current implementations of those ops in synthesis are rather ugly and unreadable as well. Therefore, after you implement the synthesis of `_basic_mult` with `net_transform`, I would like it if you could write normal PyRTL implementations of '-<>='  and use them as well as existing implementations of '+' and mux in place of the current synthesis.

abe-k:
I ran into a very strange set of issues when using `copy_block`, and I think I've finally isolated it to one: the `copy_block` code uses `Block.__iter__`, which will throw an error if there is a loop, so the tests for loop detection fail early. If I remove the check in `Block.__iter__`, I somehow get a lot of `Wires declared but not connected` from the sanity check when setting the working block in the loop detection tests.

Is it a problem if loops are detected during synthesis (i.e. that the test cases for loop detection are rewritten)? Otherwise, how to go about copying a block?

Gamrix:
[...]

PR #134, opened 2016-04-26
Synthesis of high-level operations using net_transform

abe-k:
Should resolve https://github.com/UCSBarchlab/PyRTL/issues/116

Handles synthesis of `*`, `+`, `-`, `x`, `=`, `<`, and `>` in terms of lower-level operations.
Includes new tests for synthesis of multiplexing, multiplication, and comparisons.

### DIFF
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -192,6 +192,60 @@ def _basic_mult(A, B):
     return adder_result[:result_bitwidth]
 
 
+def _one_bit_add(a, b, carry_in):
+    assert len(a) == len(b) == 1
+    sumbit = a ^ b ^ carry_in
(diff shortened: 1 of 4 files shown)


Gamrix:
Looks good.

The only comment I have is that I prefer `result = sim_trace.trace[self.output]` over `result = list(sim_trace.trace.values())[0]`

It's much easier to read.

abe-k:
The problem with trying to index into the trace is that we're post-synthesis, so the wires being simulated and traced are not the ones that were built initially.

If there was a way to get a trace by name conveniently, that could be used, but I think that's a separate issue.

Gamrix:
Ah, that's right, I forget about that. Use `block.io_map` to look up the post synthesis wire.

PR #144, opened 2016-05-11
simulation trace by name instead of identity

Gamrix:


abe-k:
I've updated the examples to trace by name. I don't know why the CI build is failing—the tests run fine for me locally, and the failures seem to have nothing to do with my changes. Any idea what's going on?

Gamrix:
PyParsing was updated on pypi 4 hours ago. It looks like as though the new version has a serious bug. I reran Travis on a commit that was committed yesterday (which passed then) and it failed as well. Lets wait until tomorrow and see if they have fixed the problem by then.

abe-k:
It looks like PyParsing 2.1.3 just got added, which presumably fixes the 2.1.2 bugs.

Gamrix:
The merge was rather hairy (as my asBlock refactor touched some of the same places) but I have gotten it in.

PR #158, opened 2016-06-29
output_to_verilog class conversion and VerilogSanitizer creation

nvandervoort:
Added `VerilogSanitizer`. Changed `output_to_verilog()` and `output_verilog_testbench()` functions to classes that use `VerilogSanitizer`. Changed code in other places in adjustment for the change to classes.


Gamrix:
Manually merged back into development

PR #164, opened 2016-06-30
Removed NameIndexer-based classes from documentation

nvandervoort:
Renamed `NameIndexer` and its children to have underscores before so they will not appear in the documentation, as they are internal classes.


Gamrix:
Manually pulled

PR #165, opened 2016-07-01
Added documentation for rtllib, estimate, corecircuits

nvandervoort:
and cleaned up docstrings in those functions, hiding some methods and clarifying others


nvandervoort:
Also fixed some formatting in the .rst files that had already been created

nvandervoort:
@Gamrix there are still param and return descriptions that should be filled out on the doc-strings for `sim_multicycle()` and `multi_sim_multicycle()`, they are marked as todo's

Issue #163, opened 2016-06-30
Fix Output to Verilog for Sync Mems


masoug:
When PyRTL outputs verilog for sync reads from mems, it uses assign statements: see here https://github.com/UCSBarchlab/PyRTL/blob/development/pyrtl/inputoutput.py#L672 for details.

Unfortunately, the synthesis tools are not smart enough to realize that the addresses for those memory reads are registers. Hence they refuse to synthesize those elements as BRAM and instead tries to convert them to memory LUTS. The solution to this was to put memory reads under an 'always' block and make the mem output wires verilog regs. Doing so does not change the semantics of the logic, it just makes it obvious to the synthesizer that the memory reads are to be done synchronously.

For more details on how the synthesis tools infer block ram, see the Xilinx design guide on page 85: http://www.xilinx.com/itp/xilinx10/books/docs/sim/sim.pdf#page86

Gamrix:
Could you post some example code?

masoug:
the design guide i linked above is the "ground truth" for Xilinx's synthesis tools. It contains sample verilog code that for-sure converts to brams.

masoug:
@Gamrix please reopen

I found another way to specify synchronous memories in the same semantic form as PyRTL. Please see https://github.com/UCSBarchlab/PyRTL/issues/199 for details. The verilog output generator needs to be changed to use blocking assignments for synchronous memory addressing.

PR #168, opened 2016-07-01
First attempt at fixing issue #163 and some other clerical changes

nvandervoort:


pyrtl/inputoutput.py
```diff
@@ -669,9 +669,12 @@ def _to_verilog_combinational(self):
                 print('    assign %s = {%s};' % t, file=self.file)
             elif net.op == 'r':
                 pass  # do nothing for registers
-            elif net.op == 'm':
+            elif net.op == 'm':  # use always block and assign as Verilog register
+                print('    always @( posedge clk )', file=self.file)
+                print('    begin', file=self.file)
                 t = (self._varname(net.dests[0]), net.op_param[0], self._varname(net.args[0]))
-                print('        assign %s = mem_%s[%s];' % t, file=self.file)
+                print('        %s <= mem_%s[%s];' % t, file=self.file)
```
masoug: did you also incorporate changes to make these destination wires verilog regs?

masoug: @Gamrix i looked at the verilog output from this and i think you need to change how the memory output registers are declared in verilog.
screenshot from 2016-07-01 16-07-06
screenshot from 2016-07-01 16-09-46
The output wires need to be declared as regs, otherwise the verilog compiler will fail.

nvandervoort:
@masoug I changed part of the header where the wire should have been a reg, but it is hard for me to tell whether that will solve the whole problem or just one instance of it.

pyrtl/inputoutput.py
nvandervoort: @masoug I changed part of the header where the wire should have been a reg, but it is hard for me to tell whether that will solve the whole problem or just one instance of it.

masoug:
screenshot from 2016-07-05 14-36-25
im getting this exception when i try to output the verilog. i guess some wirevectors may not have operations?

Gamrix:
@masoug could you post some sample code and expected output?

nvandervoort:
Okay I'll try to figure out a way for the translator to know during `_to_verilog_header()` which wires will end up being connected to mems (and therefore need to be regs)

masoug:
awesome. it works on my side of things.

Issue #167, opened 2016-07-01
Incorporate AES Encryption into the main AES file


Gamrix:
Currently, there is a working AES (PR #130). I would like it to be incorporated into PyRTL, but it to be together with the decryption (together in one file)

PR #173, opened 2016-07-07
Merged aes_encryption and tests into aes and tests for issue #167

nvandervoort:


nvandervoort:
When this is pulled, please make sure that @DivyaJanaswamy gets credit somehow, as right now it looks like I wrote a bunch of code that I just manually merged with copy and paste.

pyrtl/rtllib/aes.py
```diff
@@ -49,6 +49,11 @@ def _inv_sub_bytes(self, in_vector):
         subbed = [self.inv_sbox[byte] for byte in libutils.partition_wire(in_vector, 8)]
         return pyrtl.concat_list(subbed)
```
Gamrix: try to find a way to unify `_inv_sub_bytes()` and `_sub_bytes()`

pyrtl/rtllib/aes.py
```diff
@@ -57,18 +62,35 @@ def _inv_shift_rows(in_vector):
...
+        else:
+            return self._inv_gal_mult_dict[mult_table][c]
 
     def _inv_mix_columns(self, in_vector):
         self._build_memories_if_not_exists()
         igm_mults = [14, 9, 13, 11]
         subgroups = libutils.partition_wire(in_vector, 32)
         return pyrtl.concat_list([self._mix_col_subgroup(sg, igm_mults) for sg in subgroups])
 
+    def _mix_columns(self, in_vector):
```
Gamrix: similarly try to find a way to unify these

pyrtl/rtllib/aes.py
```diff
@@ -79,15 +101,91 @@ def _mix_single(index):
...
-        that the answer decryption result has been calculated.
+        return ready, plain_text: ready is a one bit signal showing
+        that the decryption result (plain_text) has been calculated.
         """
         if len(key_in) != len(ciphertext_in):
             raise pyrtl.PyrtlError("AES key and ciphertext should be the same length")
 
-        cipher_text, key = (pyrtl.Register(len(ciphertext_in)) for i in range(2))
+        plain_text, key = (pyrtl.Register(len(ciphertext_in)) for i in range(2))
         key_exp_in, add_round_in = (pyrtl.WireVector(len(ciphertext_in)) for i in range(2))
```
Gamrix: Umm, I don't agree with the change of cipher text to plain_text. Only at the last stage is the value in the register actually a plain_text.

pyrtl/rtllib/aes.py
```diff
@@ -308,3 +436,30 @@ def build_mem(data):
         7c 72 60 6e 44 4a 58 56 37 39 2b 25 0f 01 13 1d 47 49 5b 55 7f 71 63 6d
         d7 d9 cb c5 ef e1 f3 fd a7 a9 bb b5 9f 91 83 8d
         ''')
+
```
Gamrix: Everything after this should be moved into an example usage for the docstring at the beginning.

Gamrix:
There is only one author that can be added to a commit. Therefore, the only way you could give her credit is to add it into the commit message. You can ammend your commit messages as such:
https://help.github.com/articles/changing-a-commit-message/

It is best to change the commit message prior to making another commit. The amendment will change the commit and wipe away some of the diff notes above (when you eventually force push). Therefore, it might be the best idea to note it in the next commit you make instead.

pyrtl/rtllib/aes.py
nvandervoort: alright you're the boss, boss :ok_hand:

pyrtl/rtllib/aes.py
nvandervoort: I will have to do that in a separate commit from a separate branch because my current branch for this issue does not have the rtllib.rst file and in order to preserve indentation, etc., the usage example will have to go directly in the .rst file.

Gamrix:
So, the reason this branch cannot be automatically merged is because it was branched off of a feature branch ( #172 ) instead of off of the `development` branch. Could you use make a new branch off of a fresh copy of `upstream development` and use `git cherrypick` to put your changes into this new branch?

pyrtl/rtllib/aes.py
Gamrix: I mean that it should be part of the docstring that is inside top level of the `aes.py` file. (unless that is not a valid place to put docstrings)

nvandervoort: It is, but indentation, etc. isn't preserved in the documentation that it creates from docstrings. It would look better if it were placed in the .rst file because they have more formatting options.

Gamrix: look at http://docutils.sourceforge.net/docs/user/rst/quickref.html#literal-blocks for a way to insert a literal block. I found through looking at https://www.python.org/dev/peps/pep-0287/. The problem with sticking it in the `.rst` is that information there is much more likely to quickly go stale.

PR #176, opened 2016-07-08
Commits for issue #167

nvandervoort:


pyrtl/rtllib/aes.py
```diff
@@ -42,9 +70,11 @@ def _key_expansion(self, old_key, key_expand_round):
         x.insert(0, x[0] ^ w[0])
         return pyrtl.concat_list(x)
 
-    def _inv_sub_bytes(self, in_vector):
+    def _sub_bytes(self, in_vector, direct=0):
+        """ direct is 0 for normal, 1 for inv """
```
Gamrix: Please use booleans such as `True` and `False` instead of integers. also, i would recommend using `inverse` instead of the more vague term `direct`.

Issue #171, opened 2016-07-06
Restrict MemIndexed


Gamrix:
After a talk with Sherwood, we decided that three changes should be made to `MemIndexed`

- Reading and writing the `MemIndexed.name` should throw a descriptive error stating what 
  `MemIndexed` is. As it is a temporary object, it should not have a name
- `MemIndexed` objects should be consumed. After using them for one operation, they should 
   throw an error if a person should use them again 
   (be very careful around conditional operations!!!!!)
- `MemIndexed` should no longer be called `_MemIndexed` as we are exposing it to users

MemIndexed Name Error:
```
MemIndexed is a temporary object and therefore doesn't have a name
```

MemIndexed Reassign Error String:
```
MemIndexed can only be used once. If you are trying to use the result of a read in two different locations
please use pyrtl.aswires() to convert this MemIndexed into a WireVector. If you were intending to make
multiple ports with this MemIndexed assignment (not likely), you must explicitly make multiple ports. 
```

jemcmahan13:
I'm unclear how this consumption would work. Would the following be allowed?

```python
x = mem[addr]
y <<= x
z <<= x
```

Once it's clear the object is being used as a read, why not transparently change it into a WireVector?

Gamrix:
That code would not be allowed.
The problem with transparently changing it to a `WireVector` is that it would have to stick some additional logic between `x` and `y` on the third line (aka add logic after the fact). 

Another option is that we can transparently change `x` into a `"WireVector"` on line 2 prior to the assignment. (this could actually fix one of the other bugs that we have with memories)

~~The other option is to always have these memory reads always return a WireVector~~ Actually the way Memory writes occur means that this cannot work properly.

Gamrix:
~~Actually after thinking about it, all of the alternatives would actually require a ton of very hacky code to implement, as the MemIndexed object is not always the class getting the function call when it is accessed~~

Never mind that. There can be rather simple changes made to as_wire to accommodate this case.

timsherwood:
I am not sure how this would work, can you explain?  Also, one other crazier option is for the MemIndexed object transmute itself into a wirevector when it is read from (you can do that in python by setting self.__class__).  I am not suggesting that we should do that though!

Gamrix:
I'm thinking of the following in `as_wires(val)`
```
    elif isinstance(val, _MemIndexed):
        # covert to a memory read when the value is actually used
        if val.wire is None:
[...]

PR #180, opened 2016-07-08
Added fixes for _MemIndexed in response to issue #171

nvandervoort:
and some tests for those changes


pyrtl/memory.py
```diff
@@ -69,6 +70,10 @@ def sign_extended(self, bitwidth):
     def zero_extended(self, bitwidth):
         return as_wires(self).zero_extended(bitwidth)
 
+    @property
+    def name(self):
+        raise PyrtlError("MemIndexed is a temporary object and therefore doesn't have a name")
+
```
Gamrix: Umm, read the text at the end again. the memindexed should make accesses and changes to a _MemIndexed.name affect the underlying wire, and make the wire if needed.

Gamrix:
For the `test_write_memindexed_ior`, `test_read_memindexed_ior` and `test_read_memindexed_ilshift`, could you create a meaningful simulation (5 sim.steps for each test) to check that the circuit actually performs as intended (this will include getting rid of some of the constants)

tests/test_memblock.py
```diff
@@ -106,49 +106,81 @@ def setUp(self):
 
     def test_memindexed_name(self):
         x = self.mem[2]
-        with self.assertRaises(pyrtl.PyrtlError):
-            print(x.name)
+        x.name = 'test_name'  # setter
+        self.assertEquals(x.name, 'test_name')  # getter
+        # with self.assertRaises(pyrtl.PyrtlError):
+        #     print(x.name)
```
Gamrix: please check that the wire inside x also gets assigned `'test_name'`

Gamrix:
In order to properly check a simulation, the results of the simulation must be checked through `sim.inspect()`, `sim.inspect_mem()` or `sim_trace.trace`

Gamrix:
Also, can you add an additional unittest to make sure that doing both a memory read as well as setting of a name does not result in multiple read ports being made?

nvandervoort:
For what tests would I need to use `sim.inspect()`, etc.?

Gamrix:
the ones that are being simulated (aka the three I mentioned above).

PR #192, opened 2016-07-15
Added unit tests for area_estimation() and TimingAnalysis

nvandervoort:


Gamrix:
I don't want to run the algorithm multiple times, instead I want something like 
`self.assertIsEqual(area_estimate(), 234849)`
replace `234849` with the actual integer literal that this function returns.

https://en.wikipedia.org/wiki/Integer_literal

Issue #152, opened 2016-06-08
Use NameSanitizer in other places


Gamrix:
Recently I have built a generic `NameSanitizer` as well as base classes for PyRTL.
This is out of the work in issue #148

Places where the old code should be replaced are:
- [x] Verilog Output
- [x] Global Temporary Names (names made by normal `WireVector` instantiations)
- [x] Probe Names (use the `NameIndexer` to generate probe indexes)
- [x] RtlAssert Names (just like Probes)
- [x] VCD Output: Issue #147 (found in simulation.py)

If there are places that I have forgotten, please mention them below.

jemcmahan13:
The Name Sanitizer commit (c90d2bde683521aeeb) broke my code. The error seems to be coming from the compiled code, which makes it hard to place where something went wrong.

```
Traceback (most recent call last):
  File "main.py", line 189, in <module>
    sim.step({inval: feed_value})
  File "/home/joseph/work/PyRTL/pyrtl/simulation.py", line 379, in step
    self.regs, self.outs, mem_writes = self.sim_func(ins)
  File "<string>", line 98, in sim_func
KeyError: u'Heap-raddr'
```
(Different wires, all with names that need to be sanitized, show up as the argument of the KeyError.) Is there a place in the compiled code that uses the unsanitized names? Not sure why else it would throw a KeyError on the original name string.

Gamrix:
Yeah, the `AssertIndexer` and `ProbeIndexer` classes seem like overkill. Could we just use `NameIndexer` instances and just put the rest of the code in the `probe` and `rtl_assert` functions? the `AssertIndexer` is particularly egregious because it basically looks like a copy of `NameIndexer`.

PR #197, opened 2016-07-19
Got rid of ProbeIndexer and AssertIndexer classes for issue #152

nvandervoort:


pyrtl/helperfuncs.py
```diff
@@ -107,31 +107,8 @@ def match_bitwidth(*args):
...
-
-        p = Output(name=pname)
-        p <<= w  # late assigns len from w automatically
-        self.internal_index += 1
-        return w
-
-
-probeIndexer = _ProbeIndexer()
+probeIndexer = _NameIndexer('Probe')
+probeIndexer.internal_index += 1
```
Gamrix: get rid of the `internal_index += 1` . I don't care if the index starts at 0 or 1

pyrtl/helperfuncs.py
```diff
@@ -150,7 +127,23 @@ def probe(w, name=None):
...
+    if not isinstance(w, WireVector):
+        raise PyrtlError('Only WireVectors can be probed')
+
+    print('(%s-%d)' % (prefix, index), end=' ')
+    print(get_stack(w))
+
+    if name:
+        pname = name
+    else:
+        pname = '(%s%d__%s)' % (prefix, index, w.name)
```
Gamrix: think of a way to simplify the previous 4 lines (changing a variable name is fine)

pyrtl/helperfuncs.py
nvandervoort: Done

pyrtl/helperfuncs.py
```diff
@@ -150,7 +127,23 @@ def probe(w, name=None):
     confuse various post-processing transforms such as output to verilog)
     """
     global probeIndexer
-    return probeIndexer.next_indexed(w, name)
+    index = probeIndexer.next_index()
+    prefix = probeIndexer.internal_prefix
```
Gamrix: You don't need the `global` in this case

nvandervoort: `_NameIndexer` only has two methods and one of them is called here. The other doesn't go along with the naming convention for probes

Gamrix:
also, I would prefer that the probe name be changed back to something like this:
` pname = '(Probe-%d : %s)' % (_probe_number, w.name)`

Gamrix:
I don't care about the exact naming of the probe wire, just change the function so that it simplifies the code.

Issue #204, opened 2016-07-21
net graph throws errors when there are unconnected wires


Gamrix:
Sample code:
```
import pyrtl
a = pyrtl.WireVector(2)

# this line throws an error
pyrtl.inputoutput.net_graph()
```

PR #207, opened 2016-07-22
Fixed net_graph() unused wires problem (issue #204) and added a test for it

nvandervoort:
(made new branch to fix double-commit branch from before)


Gamrix:
Where's the aformentioned test?
Also, arg_set and dest_set are both subsets of wirevector_set.

nvandervoort:
@Gamrix I added the other commit that had the test

PR #208, opened 2016-07-22
Improved print_trace() with spacing and new parameters

nvandervoort:


pyrtl/simulation.py
```diff
@@ -793,14 +793,48 @@ def add_fast_step(self, fastsim):
...
+                file.flush()
+
+        else:
+            bfunc = dec
+            if base == 2:
+                bfunc = bin
+            elif base == 8:
+                bfunc = oct
+            elif base == 16:
+                bfunc = hex
```
Gamrix: Can you change this into a lookup table (aka a dict)?

pyrtl/simulation.py
```diff
@@ -793,14 +793,48 @@ def add_fast_step(self, fastsim):
...
+                                 for x in self.trace[w])) for w in self.trace)
+            maxlenright = (maxlenval+1) * len(list(self.trace.values())[0])
+            print(' '*(maxlenleft-1) + "--- Values in base %d ---" % base)
+            for w in sorted(self.trace, key=_trace_sort_key):
+                file.write(w.ljust(maxlenleft))
+                line = str()
+                for x in self.trace[w]:
+                    strx = bfunc(x)[(1 if bfunc == oct else 2):]
+                    padding = maxlenval - len(strx) + 1
+                    line += (' '*padding) + strx
```
Gamrix: look at https://docs.python.org/2/library/string.html#string.rjust for a slightly cleaner way of doing this

pyrtl/simulation.py
```diff
@@ -793,14 +793,48 @@ def add_fast_step(self, fastsim):
...
+        else:
+            bfunc = dec
+            if base == 2:
+                bfunc = bin
+            elif base == 8:
+                bfunc = oct
+            elif base == 16:
+                bfunc = hex
+            maxlenval = max(max((len(bfunc(x)[(1 if bfunc == oct else 2):])
+                                 for x in self.trace[w])) for w in self.trace)
```
Gamrix: The intersting thing is that you can actually remove the inside `max` and it will work (i think you'll also need to remove the both sets of parens (the ones in between `max` and `len`))

pyrtl/simulation.py
nvandervoort: So am I good?

pyrtl/simulation.py
nvandervoort: Yeah I thought so before and tried it but it wouldn't recognize `w` while in the first for loop

Gamrix: ah, that's right. To fix that problem, also flip the order of the two `for` statements.

pyrtl/simulation.py
Gamrix: no, you should be smart enough to change that into using `rjust`

nvandervoort: Right right. Can I leave the for loops expanded like that or do you want them compressed? It's easier to read and understand this way but takes a few more lines.

Gamrix: (by that I mean the lines above, (the one in the inner `for` loop)

Issue #203, opened 2016-07-20
Make print_trace more readable


Gamrix:
When print trace is printing the result of multi_digit values, there should be spaces between each individual number
Also, the values coming out of the print trace should be aligned so that they are easier to read.

Finally, there should be a way to specifiy printing a different base. When printing, also print a first line stating which base all the values are in (binary, decimal, etc.)

Gamrix:
Also, I would recommend that you right align the values
aka
```
  1
 23

not: 
 1
 23
```

nvandervoort:
@Gamrix what about the wire names? Should it look like:
```
in1     1
in2    23
out    24

**Comment by Gamrix** [comment]
yeah, wire named should be left justified

**Comment by nvandervoort** [comment]
Should these formatting changes only affect the print when the outputs are more than one digit? If not, numerous tests will break.

**Comment by Gamrix** [comment]
By default, they should add spaces for one digit numbers. I wouldn't be opposed to a optional `compact` parameter that ommits spaces, as long as it defaults to `False`.

**Comment by nvandervoort** [comment]
So you want the exact opposite of what I proposed? That will break about 18 tests in conditional and simulation...

**Comment by Gamrix** [comment]
You're welcome to add the optional parameter to the function that validates all those unittests (it shouln't be a change in all 18 places).

**Comment by nvandervoort** [comment]
But the tests in those places assume the answer *will* be compact, so setting a parameter _compact_'s default to _False_ *would* change the result in all of those places.

**Comment by Gamrix** [comment]
Yes, I am aware of that, sometimes making the correct change is painful. 18 lines of straightforward changes is not that bad.

**Comment by nvandervoort** [comment]
Okay, I was just making sure you got what I meant because you said there wouldn't be changes in 18 places. 
On another note, how does this look?
```
                --- Values in base 8 ---
(Probe-0: tmp693)    57   614   261 16010   310    22   221    14  1227   267
adder_probe         117   654   321 16050   350    62   261    54  1267   327
in1                  57    41     3   212    12     3     5     3   335     3
in2                   1    14    73    64    24     6    35     4     3    75
out                 136    30   142    20   220    44    42    30    56   156
select_probe         13     3    14     2    22     4     4     3     5    15
std_probe            57   614   261 16010   310    22   221    14  1227   267
stdout_probe         57   614   261 16010   310    22   221    14  1227   267
```

Gamrix:
I'm fine with either of the choices.
[...]

PR #210, opened 2016-07-25
Improved print_trace() with spacing and new parameters, tests

nvandervoort:
for issue #203

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -817,13 +817,11 @@ class SimulationTrace(object):
         else:
             maxlenval = max(len('{0:{1}}'.format(x, basekey))
                             for w in self.trace for x in self.trace[w])
-            maxlenright = (maxlenval+1) * len(list(self.trace.values())[0])
             file.write(' '*(maxlenleft-3) + "--- Values in base %d ---\n" % base)
             for w in sorted(self.trace, key=_trace_sort_key):
-                file.write(six.u(str(w.ljust(maxlenleft))))
-                line = ''.join('{0:{1}}'.format(x, basekey).rjust(maxlenval+1)
-                               for x in self.trace[w])
-                file.write(line.rjust(maxlenright)+'\n')
+                file.write(''.join([w.ljust(maxlenleft), ''.join('{0:{1}}'.format(x, basekey)
+                                                                 .rjust(maxlenval+1)
+                                                                 for x in self.trace[w]) + '\n']))
                 file.flush()
 
     def print_vcd(self, file=sys.stdout):
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -793,15 +793,48 @@ class SimulationTrace(object):
         for wire_name in self.trace:
             self.trace[wire_name].append(fastsim.context[wire_name])
 
-    def print_trace(self, file=sys.stdout):
-        # TODO: make multi-digit values more readable and implement better alignment (issue #203)
+    def print_trace(self, file=sys.stdout, base=10, compact=False):
+        """
+        Prints a list of wires and their current values.
+        :param int base: the base the values are to be printed in
+        :param bool compact: whether to omit spaces in output lines
+        """
         if len(self.trace) == 0:
             raise PyrtlError('error, cannot print an empty trace')
-        maxlen = max(len(w) for w in self.trace)
-        for w in sorted(self.trace, key=_trace_sort_key):
-            file.write(' '.join([w.rjust(maxlen),
-                       ''.join(str(x) for x in self.trace[w])+'\n']))
-            file.flush()
+        if base not in (2, 8, 10, 16):
+            raise PyrtlError('please choose a valid base')
+
+        def dec(i):
+            return '0d' + str(i)
+
+        maxlenleft = max(len(w) for w in self.trace)
+        if compact:
+            for w in sorted(self.trace, key=_trace_sort_key):
+                file.write(' '.join([w.rjust(maxlenleft),
+                                     ''.join(str(x) for x in self.trace[w]) + '\n']))
+                file.flush()
+
+        else:
+            bfunc = dec
+            if base == 2:
+                bfunc = bin
+            elif base == 8:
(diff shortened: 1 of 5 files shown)


nvandervoort:
@Gamrix Why won't this build...

Gamrix:
https://www.traviscistatus.com/incidents/mjyq8dprd2zx

nvandervoort:
Of course it finished building like 2 minutes after I asked...

pyrtl/simulation.py
```diff
@@ -793,15 +793,44 @@ def add_fast_step(self, fastsim):
...
+            raise PyrtlError('please choose a valid base')
+
+        def dec(i):
+            return '0d' + str(i)
+
+        def octal(i):
+            return ('0o' + oct(i)[(0 if i == 0 else 1):] if sys.version_info.major == 2
+                    else oct(i))
+
+        bfunc = {2: bin, 8: octal, 10: dec, 16: hex}[base]
```
Gamrix: I'm wondering if using `string.format` will make this code and the one below,
`line += bfunc(x)[2:].rjust(maxlenval+1) `
look simpler
https://docs.python.org/2/library/string.html#format-examples

The nice part is that with it, you don't have the problem of stripping prefixes like `0x`

pyrtl/simulation.py
```diff
@@ -793,15 +793,44 @@ def add_fast_step(self, fastsim):
...
+        else:
+            maxlenval = max(len(bfunc(x)[2:])
+                            for w in self.trace for x in self.trace[w])
+            maxlenright = (maxlenval+1) * len(list(self.trace.values())[0])
+            file.write(' '*(maxlenleft-3) + "--- Values in base %d ---\n" % base)
+            for w in sorted(self.trace, key=_trace_sort_key):
+                file.write(w.ljust(maxlenleft)+'')
+                line = str()
+                for x in self.trace[w]:
+                    line += bfunc(x)[2:].rjust(maxlenval+1)
```
Gamrix: After you simplify this (or not) using `string.format`, convert this into a `line = ''.join(...)` statement

pyrtl/simulation.py
```diff
@@ -793,15 +793,44 @@ def add_fast_step(self, fastsim):
...
+                                     ''.join(bfunc(x)[2:] for x in self.trace[w]) + '\n']))
+                file.flush()
+
+        else:
+            maxlenval = max(len(bfunc(x)[2:])
+                            for w in self.trace for x in self.trace[w])
+            maxlenright = (maxlenval+1) * len(list(self.trace.values())[0])
+            file.write(' '*(maxlenleft-3) + "--- Values in base %d ---\n" % base)
+            for w in sorted(self.trace, key=_trace_sort_key):
+                file.write(w.ljust(maxlenleft)+'')
```
Gamrix: `+ ''` does nothing in this case

nvandervoort: Actually the python 2.x tests were throwing an error saying that `write()` expected a unicode and got a string instead, so I first used the `unicode()` method but that apparently is only implemented in python 2.x, so I found that adding `+''` converts it and takes care of the 2.x problem while still working in 3.x

Gamrix: I don't see how what you did would fix anything... both of them should be a string type.
Could you do:
`print(type(w.ljust(maxlenleft)))` and `print(type(w.ljust(maxlenleft) +  ''))` for me?
also, if it is the case that you need it to be Unicode, please use:
http://pythonhosted.org/six/#six.u
Six is a python 2/3 compatability library  and should be your goto whenever you have something in Python2 that is not in python3 (or not in the same place)

nvandervoort: Yeah it said they're both type 'str' but when I took it out the `+''` the tests fail (but pass with the interpreter changed to Python 3.5). I'll use `six.u()`, thanks

pyrtl/simulation.py
nvandervoort: Wow I had no idea `format()` could do that much stuff, that will help a lot

Gamrix: I want this to be mainstream so badly (but it'll be at least 5 years...)
https://docs.python.org/dev/whatsnew/3.6.html#pep-498-formatted-string-literals

nvandervoort: That would be very useful, yes

pyrtl/simulation.py
```diff
@@ -804,31 +805,24 @@ def print_trace(self, file=sys.stdout, base=10, compact=False):
...
                             for w in self.trace for x in self.trace[w])
             maxlenright = (maxlenval+1) * len(list(self.trace.values())[0])
             file.write(' '*(maxlenleft-3) + "--- Values in base %d ---\n" % base)
             for w in sorted(self.trace, key=_trace_sort_key):
-                file.write(w.ljust(maxlenleft)+'')
-                line = str()
-                for x in self.trace[w]:
-                    line += bfunc(x)[2:].rjust(maxlenval+1)
+                file.write(six.u(str(w.ljust(maxlenleft))))
+                line = ''.join('{0:{1}}'.format(x, basekey).rjust(maxlenval+1)
```
Gamrix:
```python
Aligning the text and specifying a width:
>>>
>>> '{:<30}'.format('left aligned')
'left aligned                  '
>>> '{:>30}'.format('right aligned')
'                 right aligned'
>>> '{:^30}'.format('centered')
'           centered           '
>>> '{:*^30}'.format('centered')  # use '*' as a fill char
'***********centered***********'
```

pyrtl/simulation.py
```diff
@@ -804,31 +805,24 @@ def print_trace(self, file=sys.stdout, base=10, compact=False):
...
+                               for x in self.trace[w])
                 file.write(line.rjust(maxlenright)+'\n')
```
Gamrix: isn't the `rjust(maxlenright)` unnecessary because you are already setting a fixed width to every string in the line already (through `ljust` and `rjust`)?

nvandervoort: It would appear you are correct

Issue #195, opened 2016-07-19
Signed Integer Multiplication


Gamrix:
Implement a wrapper around Multiplication that allows for signed values to be inputted

The current multipliers in `rtllib` might  not support signed numbers.
check if they do, and find a way to add signed multiplication if it does not.

nvandervoort:
Do you want every multiplier to have signed support?

Gamrix:
I want something that can take in the Wallace or dada multiplier as an arg and return a version of those multipliers that does signed multiplication.

PR #214, opened 2016-07-27
Added signed multiplier (issue #195), two's-complement conversion funcs, tests for both

nvandervoort:


pyrtl/rtllib/multipliers.py
```diff
@@ -155,6 +155,44 @@ def tree_multiplier(A, B, reducer=adders.wallace_reducer, adder_func=adders.kogg
...
+    aneg = A[-1]
+    bneg = B[-1]
+    a = pyrtl.WireVector(len(A))
+    b = pyrtl.WireVector(len(B))
+    neg = aneg ^ bneg
+    with pyrtl.conditional_assignment:
+        with aneg:
+            a |= ~A + 1
+        with pyrtl.otherwise:
+            a |= A
```
Gamrix: You're duplicating the same 5 lines quite a lot, Could you make that a private function?

Gamrix: Make that 6 lines

nvandervoort: ^? I see the 5 you're referring to (168-172, for example)

Gamrix:
What's the sixth line that can be abstracted away?

pyrtl/rtllib/libutils.py
```diff
@@ -33,6 +33,39 @@ def str_to_int_array(string, base=16):
...
+    Takes a two's-complement represented value and
+    converts it to a signed integer based on the provided bitwidth.
+    For use with Simulation.inspect() etc. when expecting negative numbers,
+    which it does not recognize
+    """
+    valbl = val.bit_length()
+    if bitwidth < val.bit_length() or val == 2**(bitwidth-1):
+        raise pyrtl.PyrtlError("please choose a larger target bitwidth")
+    if val == 0:
+        return 0
```
Gamrix: Why does there need to be a special case for zero?

nvandervoort: That was from the flipping algorithm I was using before, I just removed it

nvandervoort:
Is the sixth line the one declaring the wires (`a`, `b`, and `res`)?

pyrtl/rtllib/multipliers.py
```diff
@@ -172,18 +171,20 @@ def signed_tree_multiplier(A, B, reducer=adders.wallace_reducer, adder_func=adde
             bits[i + j].append(x & y)
 
     res = reducer(bits, bits_length+1, adder_func)
```
Gamrix: can you call `tree_mult` instead of making a copy of the code that it had?

nvandervoort: It's not the same code, `tree_mult` uses bit_length`, not `bit_length+1`. Also it uses `_trivial_mult()` when one of the inputs is 1 bit, while the signed version is supposed to throw an error there.

Gamrix: Why do you need `bit_lengths + 1`? Also, the error is thrown earlier in the function anyways, so having the `trivial_mult` doesn't matter.

Gamrix: Also, by the first line of the copied code, we know that `a` and `b` both have a 0 in their most significant bits. shouldn't we shave those off?

nvandervoort: Idk, it doesn't work without the `+1`

nvandervoort: Oh wait I figured it out, shaved off the first bit, then multiplied, then extended the result and applied `_twos_comp_conditional()` and the result's bitwidth is `len(A) + len(B)`, which is good too

PR #215, opened 2016-07-29
Added unit tests for sanity_check_net()

nvandervoort:
in new test class `TestSanityCheckNet`

### DIFF
--- a/tests/test_core.py
+++ b/tests/test_core.py
@@ -197,6 +197,10 @@ class TestSanityCheckNet(unittest.TestCase):
         self.invalid_net("op_param out of bounds", net)
         net = self.new_net(op='s', op_param=(10,), args=1)
         self.invalid_net("op_param out of bounds", net)
+        net = self.new_net(op='s', op_param=(0, 1, 3), args=1)
+        self.invalid_net("op_param out of bounds", net)
+        net = self.new_net(op='s', op_param=(0, True, False, dict(), 'hi'), args=1)
+        self.invalid_net("select op_param requires ints", net)
 
     def test_net_wrong_op_param_mem(self):
         for op in 'm@':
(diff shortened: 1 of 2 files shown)


tests/test_core.py
```diff
@@ -96,6 +86,166 @@ def test_block_iterators(self):
...
+                           args=tuple(pyrtl.Input(i) for i in (1, 2, 2)))
+        self.invalid_net("op_param should be None", net)
+
+    def test_net_wrong_op_param_select(self):
+        net = self.new_net(op='s', op_param='hi', args=1)
+        self.invalid_net("select op requires tuple op_param", net)
+        net = self.new_net(op='s', op_param=(-2,), args=1)
+        self.invalid_net("op_param out of bounds", net)
+        net = self.new_net(op='s', op_param=(10,), args=1)
+        self.invalid_net("op_param out of bounds", net)
```
Gamrix: could we have an test that tests tuples with non-int values such as `True`, and also a test to make sure that subsequent values in a tuple throw errors even if the initial one is valid?

nvandervoort: Do you mean new unit tests in `test_net_wrong_num_op_params()` or new checks in `sanity_check_net()`?

tests/test_core.py
```diff
@@ -96,6 +86,166 @@ def test_block_iterators(self):
...
+        const = pyrtl.Const(2)
+        outp = pyrtl.Output(2)
+        net = self.new_net(dests=(inp,))
+        self.invalid_net("Inputs, Consts cannot be destinations", net)
+        net = self.new_net(dests=(const,))
+        self.invalid_net("Inputs, Consts cannot be destinations", net)
+        net = self.new_net(args=(outp, outp))
+        self.invalid_net("Outputs cannot be arguments", net)
+
+        wrong_ops = ('%', '!', 'a', 'f', '<<', '>>', '&&', '||', '==')
```
nvandervoort: test

tests/test_core.py
nvandervoort: I ended up adding a test that starts out in bounds and then goes out (as per your second request) and added a check in `sanity_check_net()` that checks whether each op_param is an int (and a test for that).

Issue #202, opened 2016-07-20
Sanity Check Unittests


Gamrix:
We need some more unittests to make sure that sanity check is working as intended.

Gamrix:
Part 1: `sanity_check_net`:
- [ ] convert current checks into `unittest.assertRaisesRegExp` checks
- [ ] general checks and `Input`/`Output` checks
- [ ] net arguments checks
- [ ] net op-param checks
- [ ] net dests checks

PR #224, opened 2016-08-30
Added tests to TestSanityCheck in response to issue #202

nvandervoort:


tests/test_core.py
```diff
@@ -426,8 +426,42 @@ def test_as_graph_duplicate_args(self):
...
+        out <<= inp
+        self.sanity_error("Duplicate wire names")
+
+    def test_unknown_wires(self):
+        inp = pyrtl.Input(8, 'inp')
+        out = pyrtl.Output(8, 'out')
+        out <<= inp
+        pyrtl.working_block().wirevector_set.discard(inp)
+        with self.assertRaises(pyrtl.PyrtlInternalError):  # sanity_check_net()
+            self.sanity_error("Unknown wires")
```
Gamrix: Wait, are two errors being thrown here?

nvandervoort: Um yeah `sanity_check_net()` throws one and then later on in `sanity_check()` another one is thrown.

PR #235, opened 2018-05-02
WaveDrom renders in Jupyter

kmahorker:
iPynb typed files can now display output in the WaveDrom format. See: http://wavedrom.com/

### DIFF
--- a/pyrtl/inputoutput.py
+++ b/pyrtl/inputoutput.py
@@ -419,6 +419,7 @@ def trace_to_html(simtrace, trace_list=None, sortkey=None):
 
     wave_template = (
         """\
+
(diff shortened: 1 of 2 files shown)


Gamrix:
I'm confused. What was the problem before?

kmahorker:
There was initially some code to output with javascript when displaying within an ipynb file, and the error was that "w in tracelist had no attribute name." The tracelist seemed to contain a strings of keys, so w is actually the name itself. By removing the ".name" the bug was fixed and the WaveDrom scripts render the output properly.

kmahorker:
Added a UnitTest to test_inputoutput.py to see if trace_html function compiles. Code Coverage issue is now solved.

PR #252, opened 2018-11-20
toFirrtl code

shannon1006:


timsherwood:
Hi Shannon, it looks like most of the tests pass, but it needs to pass "pep8" as well.  To run all the checks use the command "tox" but to just run the pep8 checks use "tox -e pep8".  Most of them are that your line length is too long.

PR #269, opened 2020-02-27
Adds simulation section for example fibonacci implementation; adds simulation helper function

mdko:


timsherwood:
This is a cool addition, thanks!

PR #277, opened 2020-04-09
Add ability to get handle to internally-defined memories (for use in Simulation) (Version 2)

mdko:
This adds the ability to get a reference to a memory created internally to any function/block.

This useful for when a block defines its own internal memory block, and during simulation you want to instantiate that memory with certain values for testing. Since the Simulation constructor requires a reference to the memory object itself, but the block your testing defines the memory internally, it's not possible get it without making that local variable an attribute of the function. This decorator does this **automagically** for you.

This version uses Python introspection in order to work. [Another pull request](https://github.com/UCSBarchlab/PyRTL/pull/278) manipulates the working block in order to store and get these memory block references. This pull request is intended to be a discussion on 1) if this is useful (I think it is) and 2) which way (this pull request or [that one](https://github.com/UCSBarchlab/PyRTL/pull/278)) is a better approach.

The one advantage is that you don't need to name your internal memories to access them (unlike the other pull request), so no need to mess with internal block code.
        
            import pyrtl
            from pyrtl.simulation import external

            @external('mem')   # <======= New stuff
            def special_memory(read_addr, write_addr, data, wen):
                mem = pyrtl.MemBlock(bitwidth=32, addrwidth=5)
                mem[write_addr] <<= pyrtl.MemBlock.EnabledWrite(data, wen & (write_addr > 0))
                return mem[read_addr]

            read_addr = pyrtl.Input(5, 'read_addr')
            write_addr = pyrtl.Input(5, 'write_addr')
            data = pyrtl.Input(32, 'data')
            wen = pyrtl.Input(1, 'wen')
            res = pyrtl.Output(32, 'res')
            
            res <<= special_memory(read_addr, write_addr, data, wen)
            sim = pyrtl.Simulation(memory_value_map={
                special_memory.mem: {  # <===== New stuff
                    0: 5,
                    1: 6,
                    2: 7,
                }
            })

            inputs = {
                'read_addr':  '012012',
                'write_addr': '012012',
                'data':       '890333',
                'wen':        '111000',
            }
            expected = {
                'res': '567590',
            }
            sim.step_multiple(inputs, expected)

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -1214,6 +1214,7 @@ class SimulationTrace(object):
         if extra_line:
             print(file=file)
 
+
 def _call_function_get_frame(func, *args, **kwargs):
     """
     Calls the function *func* with the specified arguments and keyword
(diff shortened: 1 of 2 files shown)


timsherwood:
Hi Michael, I responded in the other pull request I could see a decorator capturing a convenience aspect of this but I would prefer to avoid it it being the primary way to access some information from the netlist. I key principle in pyrtl is as simple as possible (but no simpler), and I think an access function in core to facilitate this type of behavior would be the preferred method.

PR #278, opened 2020-04-09
Add ability to get handle to internally-defined memories (for use in Simulation) (Version 1)

mdko:
This adds the ability to get a reference to a memory created internally to any function/block (as long as you have its name)

This useful for when a block defines its own internal memory block, and during simulation you want to instantiate that memory with certain values for testing. Since the Simulation constructor requires a reference to the memory object itself, but the block your testing defines the memory internally, this allows you to get the object reference.

Note that this requires you know the name of the memory block, meaning that you most likely need to have named it yourself.

This version manipulates the working block in order to store and get these memory block references. [Another pull request](https://github.com/UCSBarchlab/PyRTL/pull/277) uses Python introspection and a decorator instead and thus doesn't mess with any of the internals of PyRTL. This pull request is intended to be a discussion on 1) if this is useful (I think it is) and 2) which way (this pull request or [that one](https://github.com/UCSBarchlab/PyRTL/pull/277)) is a better approach.
        
           def special_memory(read_addr, write_addr, data, wen):
                mem = pyrtl.MemBlock(bitwidth=32, addrwidth=5, name='special_mem')
                mem[write_addr] <<= pyrtl.MemBlock.EnabledWrite(data, wen & (write_addr > 0))
                return mem[read_addr]

            read_addr = pyrtl.Input(5, 'read_addr')
            write_addr = pyrtl.Input(5, 'write_addr')
            data = pyrtl.Input(32, 'data')
            wen = pyrtl.Input(1, 'wen')
            res = pyrtl.Output(32, 'res')

            res <<= special_memory(read_addr, write_addr, data, wen)

            # Can only access it after the `special_memory` block has been instantiated/called
            # New stuff =====V
            special_mem = pyrtl.working_block().get_mem_block_by_name('special_mem')

            sim = pyrtl.Simulation(memory_value_map={
                special_mem: {
                    0: 5,
                    1: 6,
                    2: 7,
                }
            })

            inputs = {
                'read_addr':  '012012',
                'write_addr': '012012',
                'data':       '890333',
                'wen':        '111000',
            }
            expected = {
                'res': '567590',
            }
            sim.step_multiple(inputs, expected)

### DIFF
--- a/pyrtl/core.py
+++ b/pyrtl/core.py
@@ -262,7 +262,7 @@ class Block(object):
         # pre-synthesis wirevectors to post-synthesis vectors
         self.legal_ops = set('w~&|^n+-*<>=xcsrm@')  # set of legal OPS
         self.rtl_assert_dict = {}   # map from wirevectors -> exceptions, used by rtl_assert
-        self.memblocks_by_name = {}  # map from name->memblock, for easy access to memblock objs
+        self.memblock_by_name = {}  # map from name->memblock, for easy access to memblock objs
 
     def __str__(self):
         """String form has one LogicNet per line."""
@@ -307,7 +307,7 @@ class Block(object):
         (e.g. for instantiating during a simulation).
         """
         self.sanity_check_memblock(mem)
-        self.memblocks_by_name[mem.name] = mem
+        self.memblock_by_name[mem.name] = mem
 
     def get_memblock_by_name(self, name, strict=False):
         """ Get a reference to a memory stored in this block by name.
@@ -362,8 +362,8 @@ class Block(object):
             }
             sim.step_multiple(inputs, expected)
         """
-        if name in self.memblocks_by_name:
-            return self.memblocks_by_name[name]
+        if name in self.memblock_by_name:
+            return self.memblock_by_name[name]
         elif strict:
             raise PyrtlError('error, block does not have a memblock named %s' % name)
         else:
--- a/pyrtl/core.py
+++ b/pyrtl/core.py
@@ -262,7 +262,7 @@ class Block(object):
         # pre-synthesis wirevectors to post-synthesis vectors
         self.legal_ops = set('w~&|^n+-*<>=xcsrm@')  # set of legal OPS
         self.rtl_assert_dict = {}   # map from wirevectors -> exceptions, used by rtl_assert
-        self.mem_blocks = {}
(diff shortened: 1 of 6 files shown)


timsherwood:
Thanks Michael, I agree this is addressing a real need and should be supported by PyRTL.    Of the two I lean more towards this one in keeping with the "explicit is better than implicit" mindset.   I could see a decorator capturing a convenience aspect of this but I would prefer to avoid it it being the primary way to access some information from the netlist.  Decorators are also not that beginner friendly which is another consider.

I like the "get_memblock_by_name" approach in that it mirrors an existing and well used function already in "core", although I need to think if that mirroring as accurate (since a memblock can actually be referenced by multiple operators in the netlist).  Another options would be a function that returns  *all* memory blocks referenced.  That might be another option and look something like this (functions names TBD)

```
        m = pyrtl.working_block().all_the_referenced_memblocks() 

        sim = pyrtl.Simulation(memory_value_map={
            m['special_mem']: {
                0: 5,
                1: 6,
                2: 7,
            },
            m['other_special_mem']: {
                ...
            }
        })
```

timsherwood:
BTW, in thinking about it further if we go this way it means we probably need "get_mem_block_by_name" too, although I would prefer "get_memblock_by_name" just to be consistent with names (since it is "get_wirevector_by_name").  If you make this name change and we have tests in place for this new core functionality I will merge this pull request.  Sorry it took a while to get back to you on this!

mdko:
Regarding the suggestion for a type of `all_the_referenced_memblocks()` method:

I've updated the code to have a more consistent naming scheme as with wirevectors (i.e. `get_memblock_by_name()`, and I noticed that since the underlying dict storing the name-to-memblock mapping is accessible, do we really need a separate method, or can the user just access the map like so:

       m = pyrtl.working_block().memblock_by_name

       sim = pyrtl.Simulation(memory_value_map={
            m['special_mem']: {
                0: 5,
                1: 6,
                2: 7,
            },
            m['other_special_mem']: {
                ...
            }
        })

This is similar to how getting the wirevectors works (just use `pyrtl.working_block().wirevector_by_name`). If you're okay with the above, the latest set of commits should suffice.

timsherwood:
Looks good, merged.  Thanks Michael this is a nice improvement!

PR #279, opened 2020-05-06
Fix Issue with Merging BLIF Output WireVectors

mdko:
When importing BLIF models into PyRTL, there is a bug where 1-bit BLIF output wires are combined in reverse order to create an N-bit PyRTL Output WireVector. An example is in order.

Given the Verilog code:
```
module counter (clk, rst, en, count);

    input clk, rst, en;
    output reg [3:0] count;

    always @(posedge clk)
        if (rst)
            count <= 4'd0;
        else if (en)
            count <= count + 4'd1;

endmodule
```

which generates the following BLIF (first few lines shown only):
```
.model counter
.inputs clk rst en
.outputs count[0] count[1] count[2] count[3]
```

the current usage of `pyrtl.input_from_blif(f)` will create a 4-bit `count` WireVector in PyRTL which __should__ be the concatenation of `count[3]`, `count[2]`, `count[1]`, and `count[0]`, where `count[3]` is the MSB and `count[0]` is the LSB. But currently it concatenates them in reverse order, such that `count[3]` is the LSB and `count[0]` is the MSB.

This pull request simply [changes](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:blif-import?expand=1#diff-c18524de449e49323ee4ea311060d299R134) the function used to concatenate them. It also adds and improves some related tests and fixes some pep8 formatting issues.

Side note: I'm pretty sure the generating of Input WireVectors is fine, based on visual inspection.
Side note #2: I'm also pretty sure the `merge_io_vector` argument to `input_from_blif()`, when False, does not do what we expect it to do. TLDR is that it will only create one of the N 1-bit wirevectors as Input/Output, and fixing this may be nontrivial since we don't allow Output wirevectors to be arguments to nets. This issue with the flag will be fixed in a separate pull request.


mdko:
One other thing I just remembered: there is currently an issue in the BLIF parsing where it doesn't expect/allow there to be an initial value specified for the `.latch` command. The BLIF manual says that the format of `.latch` is:

`.latch <input> <output> [<type> <control>] [<init-val>]`

Where `<init-val>` can be 0, 1, 2 (for don't care), and 3 (for unknown).

For the BLIF design I imported, I manually removed the initial value and the design worked fine. Before fixing it (in one of two ways), I need to understand the implications of fix option #1 (just ignoring the init-value passed in the BLIF file) or fix option #2 (integrating that initial-value in the generated PyRTL).

Or perhaps there are some assumptions I'm missing, and we assume the BLIF latches should always be don't cares? Or something else?

timsherwood:
I am merging the BLIF fix (assuming it has been tested?).  

Re: initial values -- PyRTL does not include or manage any explicit reset logic (it is added during verilog generation) so that either async or sync reset is possible depending on how the logic might be used.  During simulation we always assume all stateful elements (e.g. registers) come up in a zero state.  It should be safe to ignore init-val 0, 2, and 3 on import.  init-val 1 should either be an error OR perhaps the net could be the complement of a register.

mdko:
Regarding testing: Yes, the commits included added and expanded the BLIF-related tests in `tests/test_inputoutput.py`, and also ran it end-to-end, for example [here](https://github.com/pllab/pl-systems-research/blob/master/hdls/PyRTL-Experiments/verilog/counter_import.py).

I will make a pull request for the init-val fix (either error or the complement net approach for init-val 1) later.

PR #280, opened 2020-05-06
Minor fixes to Verilog testbench output

mdko:
Fixes bug where internal names in Verilog design under test don't match the names referenced by the test bench. Also adds the ability to automatically include a file in the test bench (with the intention being this file contains the top level module being tested by the test bench).

No unittests were added, but I was able to test this creating and testing Verilog in my own PyRTL-to-Verilog workflow.

### DIFF
--- a/pyrtl/verilog.py
+++ b/pyrtl/verilog.py
@@ -26,7 +26,7 @@ def output_to_verilog(dest_file, block=None):
 
     block = working_block(block)
     file = dest_file
-    internal_names = _VerilogSanitizer('_verout_tmp_')
+    internal_names = _VerilogSanitizer('_ver_out_tmp_')
 
     for wire in block.wirevector_set:
         internal_names.make_valid_string(wire.name)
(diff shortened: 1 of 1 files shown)


timsherwood:
Thanks!  Testing of verilog import/export has been an issue from the start -- it would be neat if there was a way we could run Icarus or something to check the validity and correctness of the generated verilog code.

mdko:
Yes, I noticed that the testbench sets input assignments on each cycle based on the simulation trace, but doesn't include checks against expected values. We may be able to leverage the `step_multiple` function, which already takes in as an argument expected values at each step, into the generated Verilog, perhaps by adding `if (output != expected) $display("some failure message");` and run it with icarus like you say. Or perhaps hook in a model checker or use yosys' model-checking abilities afterwards.

timsherwood:
is not merged?

timsherwood:
is merged now.  sorry :)

PR #281, opened 2020-05-08
Fix unconnected input/output/consts when outputting to graphviz

mdko:
As I was trying to understand the graph structure of some circuits, I noticed that inputs/outputs/consts weren't connected to the graph. In addition, it was a little difficult to understand what some of the nodes do (select, buf), so I changed some shapes, colors, and names to make it easier to read.

The bug fixed ([Line 419](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:graphviz?expand=1#diff-c18524de449e49323ee4ea311060d299R419)) is connecting the inputs/outputs/consts; the rest are cosmetic changes that can be removed if they prove too controversial.

Given this code:

```python
import pyrtl

i = pyrtl.Input(8, 'i')
a = pyrtl.Register(8, 'a')
b = pyrtl.Output(8, 'b')

a.next <<= a + i + 2
b <<= a

sim = pyrtl.Simulation()
sim.step_multiple({'i': [1,2,3,4,5]}, nsteps=5)

sim.tracer.render_trace()

with open("ioreg.dot", "w+") as f:
    pyrtl.inputoutput.output_to_graphviz(f)
```

Before it produced the following dot file:
graph-before

Now it produces:
graph-after

Here's a larger, separate example with more node types:
gcd

### DIFF
--- a/pyrtl/inputoutput.py
+++ b/pyrtl/inputoutput.py
@@ -416,7 +416,7 @@ def net_graph(block=None, split_state=False):
             graph[w] = {}
 
     # add all of the edges
-    for w in (dest_set & arg_set):
+    for w in (dest_set | arg_set):
         try:
             _from = wire_src_dict[w]
         except Exception:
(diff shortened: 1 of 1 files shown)


mdko:
Only some of the shapes have been changed to look like the traditional symbol of their underlying gate because of the lack of symbols in the Graphviz library.

timsherwood:
yes please!!!

timsherwood:
We should get that picture on the website.

PR #282, opened 2020-05-09
Allow BLIF .latch init-val of 0, 2, or 3; blif import function takes in clock name

mdko:
An issue that was left outstanding from pull request https://github.com/UCSBarchlab/PyRTL/pull/279: BLIF files with init-vals for .latch wouldn't be accepted by the PyRTL blif parser. This fix allows the init-val to be 0, 2, or 3. I added a unit test for this.

In addition, because the clock input to modules is considered special while parsing, I thought it made sense for the user to be able to specify exactly the name of the clock their top module uses. For example, when I was importing some BLIF files, they were using a block named 'clk_i', for example. This allows you to specify the clock name via `pyrtl.inputoutput.input_from_blif(bliffile, clock_name="clk_1"). This is helpful because then the clock won't be considered an Input wirevector and PyRTL won't complain that you didn't drive it during simulation. I tested this while import BLIF files.


timsherwood:
Thanks, this definitely improves the quality of life for BLIF users

PR #291, opened 2020-09-07
added matrices to PyRTL

DKupshM:
This pull request adds matrices to PyRTL. Specifically, adding a matrix file that includes a Matrix class.

The Matrices support the following operations:
* Add
* Subtract
* Element-wise or Scalar Multiply
* Matrix Multiply
* Matrix Power
* Transpose
* Reverse

Additionally, there is also the following functions:
* Summation
* Minimum
* Maximum
* Argmax
* Dot

The Matrix class works by using a 2d list/array of WireVectors.

This should be helpful class for those working with arrays. I think a future update should increase support for signed numbers and allow multi-cycle operations.


timsherwood:
Hi! 

This is looking really good.  A couple of things to update before I think it is ready to merge:
1) I was thinking that this should probably go under rtllib/matrix.py which is where we keep the core pyrtl library hardware (of which I think matrix multiply would probably fit
2) It would probably be good to discuss what goes into the pyrtl rootlevel namespace because I am little worried about confusion from summation, minimum, maximum, argmax, dot, etc.   In general, the more we can get it to look close to NumPy the easier it will be to get people to use it :)
3) Any non-research PyRTL code needs to be covered by an extensive set of tests.  Ideally it should have at least 95% coverage (and in general 100% really the target).  For a dynamic language like Python that is really the only way to see when things go sideways when other additions are made to the codebase later on.

This is looking great though and I am excited to bring this to development release ASAP!

DKupshM:
I have renamed the function to their NumPy equivalents. I originally changed them to avoid overriding the builtin python functions.

I also moved everything to rtllib/matrix.py and removed it from the root-level namespace. I see how that can easily get confusing and I think it fits better in the scope of the project. Another possibility is keeping the base matrix class in the root-level namespace, while the matrix functions (max, min, sum, etc) are in rtllib.

For the testing, I hopefully fixed the tests not running. Locally, I have 100% code coverage. Travis ci had problems with backwards compatibility (matrix multiplication operation was added in python 3.5)  and NumPy. For the Numpy issue, I currently have the tests comparing the Pyrtl Matrix version operations against the NumPy equivalents. However, this would introduce a new root-level dependency, which is against the guidelines. I'm wondering if adding NumPy as a dependency is a good idea or if I should write my own versions of the NumPy functions to test against.

timsherwood:
Awesome, I must have missed the tests in my first review!  Regarding the tox problems with 2.7, that is indeed an issue for now.  I am going to keep 2.7 support until 2.7 is no longer the default python install for apple just because easy one-shot install for newbies is one of the goals of PyRTL.  For your tests I would suggest calling __matmul__ directly (which avoids the new parsing in 3.5 that makes '@' an operator) and/or removing certain examples for 2.7 using a test on [six.PY3](https://six.readthedocs.io/).  It is important that tox pass with no errors so that Travis-CI does not kick any new updates back.

DKupshM:
I removed the @ operator instead just using the methods. It should be python 2.7 compliant.

I think the remaining question is whether numpy can be added as a dependency or I should write my own versions of the functions to test against or some other solutions. Do you have any suggestions?

DKupshM:
Fixed tests working and code style.

mdko:
Merged in series of commits ending in e9d46634f6c9549fcf1e59005e3f2bf5bf1805ea

PR #295, opened 2020-10-03
Fixing typos, pep8, pylint, and some 2.7 test compat issues

mdko:
Fixes various typos and more importantly some tox issues (pep8, pylint, and some 2.7 test compatibility issues by adding usage of `six` where appropriate). The 2.7 compatibility issues fixed here weren't syntactic (so the code in the current development branch still runs for 2.7), but the additional fixes here make the tests more robust to reflect when 2.7 is being run so the correct exception is caught, and uses `six.callable` rather than `type` in the Bundle `get_fields` method.


timsherwood:
Thanks!  Yeah, I realized that travis CI dropped us for some reason.  Need to get CI going again

PR #321, opened 2021-02-26
Adding option to name constants

mdko:
This change officially allows constants to be given user-defined names.

The reasons why this might be desired:

* It's consistent with how we can name all other WireVectors and Memories
* Constants whose name have significant meaning (like archid, vendorid in the case of a RISC-V processor) will be reflected in various forms of output like Verilog and graphviz
* By being able to name constants explicitly, they will be automatically added to the traces (unless otherwise specified), again consistent with other WireVectors

One difference that might trip people up between naming constants and naming other wirevectors is that the name argument for constants is third, which I think is most natural (since the first argument is the constant value and the second is the bitwidth). The name follows the bitwidth like all other WireVector initializers, so it's consistent in that respect:

```python
pyrtl.Const(42, 16, "mySpecialConst")
# compared to
pyrtl.WireVector(16, "myWire")
pyrtl.Register(16, "myReg")
# etc.
```

It also modifies how named constants are presented in svg/graphviz output.

For example, given:
```python
import pyrtl

i = pyrtl.Const(0b1100)
j = pyrtl.Const(0b011, bitwidth=3)
k = pyrtl.Const(0b100110, name='k')
o = pyrtl.Output(13, 'o')
o <<= pyrtl.concat(i, j, k)

with open("consts.svg", "w") as f:
    pyrtl.output_to_svg(f)

with open("consts.gv", "w") as f:
    pyrtl.output_to_graphviz(f)
```

It produces:
consts

### DIFF
--- a/pyrtl/inputoutput.py
+++ b/pyrtl/inputoutput.py
@@ -558,7 +558,8 @@ def _default_node_namer(node, split_state=False, extra_node_info=None):
         return v
 
     if isinstance(node, Const):
-        return '[label="%d", shape=circle, fillcolor=lightgrey]' % label(node.val)
+        name = node.name + ': ' if not node.name.startswith('const_') else ''
+        return '[label="%s", shape=circle, fillcolor=lightgrey]' % label(name + str(node.val))
     elif isinstance(node, Input):
         return '[label="%s", shape=invhouse, fillcolor=coral]' % label(node.name)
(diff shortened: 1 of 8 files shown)


timsherwood:
This is great, after sleeping on it I am totally sold that this is the right way to go.  Merged!

PR #340, opened 2021-03-24
Update blif importer to allow multi-model import

mdko:
This change allows BLIF files with multiple modules/submodules ("subckt"s) to be imported. The main gist is that each submodule instantiation keeps track of a mapping between the given textual wire name and the actual wire (with an internal name) created. This way name clashes are avoided for submodules that are instantiated multiple times. A happy result is that a latent bug was discovered and fixed as well (when merge_io_vectors is False and a single-bit top-level output exists, there was ambiguity before; this indirection for preventing name clashes also fixed this).

Each submodule reference creates a copy of that module's logic; there is no reuse, which would be more efficient, but would definitely break the current PyRTL abstractions/require a lot more plumbing.


mdko:
I still need to deal with the clock; we don't expose the clock as an input, but it is possible that submodules take in that clock input and want to use it.

PR #341, opened 2021-03-24
Update blif importer to allow multi-model import

mdko:
This change allows BLIF files with multiple modules/submodules ("subckt"s) to be imported. The main gist is that each submodule instantiation keeps track of a mapping between the given textual wire name and the actual wire (with an internal name) created. This way name clashes are avoided for submodules that are instantiated multiple times. A happy result is that a latent bug was discovered and fixed as well (when merge_io_vectors is False and a single-bit top-level output exists, there was ambiguity before; this indirection for preventing name clashes also fixed this). It also fixes a bug in the allowable signal names (adds a dash).

Each submodule reference creates a copy of that module's logic; there is no reuse, which would be more efficient, but would definitely break the current PyRTL abstractions/require a lot more plumbing. It does not use any of the Module class from that other branch; it all still imports into plain PyRTL.

It also handles when submodules have clock parameters, by ensuring those inputs aren't created.


mdko:
One thing still needs to be handled (or ignored if we want to add some more stipulations) is when the _output_ of a subcircuit is used as a clock. Right now only input clocks that transitively originate from the top-level clock are tracked, but not clocks derived via some logic involving other clocks or wires (i.e. it is only expected that clocks are used for the rising edge of `.latch` elements).

PR #343, opened 2021-04-17
Improve how bit selects and memory ops are presented in graphviz

mdko:
This PR improves how bit selects are presented in graphviz.

For example, given the following code:
```python
import pyrtl

a = pyrtl.Input(32, 'a')
b = pyrtl.Output(32, 'b')
c = pyrtl.Output(5, 'c')
d = pyrtl.Output(4, 'd')
w = a[0:20:3]  # slice with custom step
b <<= pyrtl.mux(0, w, 1)
c <<= w[2:6] & (a[0] ^ 1)
d <<= w[4:6] ^ 4

with open("test1.svg", "w") as f:
    pyrtl.output_to_svg(f)
```

Previously we had:
test1

And now we have:
test1

For non-consecutive selected bits, the word 'bits' is displayed. Other choices we can change that to in the future are 'slice' or 'select', for example.

It also does a small improvement on memory display, by including the user-defined name if present.

Before:
test2

After:
test2-2

### DIFF
--- a/pyrtl/inputoutput.py
+++ b/pyrtl/inputoutput.py
@@ -760,13 +760,20 @@ def _default_node_namer(node, split_state=False, extra_node_info=None):
             elif node.op == 'x':
                 return '[label="%s", shape=invtrapezium]' % label("mux")
             elif node.op == 's':
-                selEnd = node.op_param[0]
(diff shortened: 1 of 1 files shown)


zsisco:
🙌

PR #344, opened 2021-04-17
Put all io-related code into new inputoutput directory, splitting up original file as needed

mdko:
It was getting unwieldy putting all the IO-related code (SVG output, BLIF input, FIRRTL output, etc.) into one single file (`inputoutput.py`). Since the Verilog-related code was already in its own file, I think it makes sense to separate out the other major portions into their own files (`blif.py`, `firrtl.py`, and `graphs.py`), and put them all in a subdirectory called `inputoutput`. Everything in them is still accessible via `pyrtl.`, e.g. `pyrtl.output_to_verilog` or `pyrtl.input_from_blif`. This also makes it a little cleaner when we want to add things like `coreir.py`, `llhd.py`, etc., or export BLIF by adding it to `blif.py`.


mdko:
Dropping PR based on offline discussion: we should try and avoid subdirectories in `pyrtl/` when possible.

PR #347, opened 2021-04-23
Remove `replace_wire`, use `replace_wire_fast` instead; add tests and more documentation

mdko:
`replace_wire` is incorrect, I believe. In [line 101](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:transform-cleanup?expand=1#diff-67c9ebb170eeb38059bfed23e32ad4d0f03cba3f757e8f2085c6e768ece6302fL101), it uses `new_src` when it should be using `new_dst`. In addition, iterating over `block.logic` in lines [85](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:transform-cleanup?expand=1#diff-67c9ebb170eeb38059bfed23e32ad4d0f03cba3f757e8f2085c6e768ece6302fL85) and [96](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:transform-cleanup?expand=1#diff-67c9ebb170eeb38059bfed23e32ad4d0f03cba3f757e8f2085c6e768ece6302fL96) causes alterations to `block.logic` in lines [92](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:transform-cleanup?expand=1#diff-67c9ebb170eeb38059bfed23e32ad4d0f03cba3f757e8f2085c6e768ece6302fL92) and [103](https://github.com/UCSBarchlab/PyRTL/compare/development...pllab:transform-cleanup?expand=1#diff-67c9ebb170eeb38059bfed23e32ad4d0f03cba3f757e8f2085c6e768ece6302fL103), respectively, to cause a "Set changed size during iteration" error.

In addition, this was only used in `wire_transform`, which has a single unit test. I think we can instead just use and export `replace_wire_fast` and `replace_wires`, and use `replace_wire_fast` where `replace_wire` was used in `wire_transform`.

### DIFF
--- a/pyrtl/__init__.py
+++ b/pyrtl/__init__.py
@@ -110,4 +110,9 @@ from .passes import optimize
 from .passes import one_bit_selects
 from .passes import two_way_concat
 
-from .transform import net_transform, wire_transform, replace_wire, copy_block, clone_wire
+from .transform import net_transform
+from .transform import wire_transform
+from .transform import copy_block
+from .transform import clone_wire
+from .transform import replace_wires
+from .transform import replace_wire_fast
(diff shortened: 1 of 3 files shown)


mdko:
Question: should we also make a wrapper function, `replace_wire(orig_wire, new_wire, block=None)`, that just calls`replace_wires({orig_wire: new_wire}, block=block)`?

PR #390, opened 2021-06-15
Add ability to specify how values render in trace; improve symbol_len calculation.

mdko:
As I've been analyzing many traces, I've found it useful to be able to specify the format that the values are represented in the wave form. I.e. I'd like to be able to specify that the numbers should be represented as hexadecimal (the default), or binary, or octal, or decimal, or some other format. This PR makes it possible by passing in a representation function to `render_trace()`. This also adds more tests.

E.g. now we get (here is the same trace printed in binary, hex, oct, and decimal one after the other), via this code:
```python
sim.tracer.render_trace(repr_func=bin, symbol_len=None)
sim.tracer.render_trace(repr_func=hex, symbol_len=None)
sim.tracer.render_trace(repr_func=oct, symbol_len=None)
sim.tracer.render_trace(repr_func=str, symbol_len=None)
```

Screen Shot 2021-06-15 at 11 28 08 AM

It also works for the `trace_to_html` (used by the IPython notebooks):

Screen Shot 2021-06-15 at 4 16 32 AM

Screen Shot 2021-06-15 at 4 16 49 AM

Screen Shot 2021-06-15 at 4 17 10 AM

Screen Shot 2021-06-15 at 4 17 22 AM

Also adds a `repr_per_name` function that allows you to specify a representation per wire, via a dictionary mapping a wire name to a function that produces a string. This way, you get things like:

Screen Shot 2021-06-15 at 2 13 54 PM

Screen Shot 2021-06-15 at 2 13 39 PM

(ability to see the names of enums used).

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -876,21 +876,23 @@ class _WaveRendererBase(object):
         num_tick = self._tick + str(n)
         return num_tick.ljust(symbol_len * segment_size)
 
-    def render_val(self, w, n, current_val, symbol_len):
+    def render_val(self, w, n, current_val, symbol_len, repr_func):
         if w is not self.prev_wire:
             self.prev_wire = w
             self.prior_val = current_val
-        out = self._render_val_with_prev(w, n, current_val, symbol_len)
+        out = self._render_val_with_prev(w, n, current_val, symbol_len, repr_func)
         self.prior_val = current_val
         return out
 
-    def _render_val_with_prev(self, w, n, current_val, symbol_len):
+    def _render_val_with_prev(self, w, n, current_val, symbol_len, repr_func):
         """Return a string encoding the given value in a waveform.
 
         :param w: The WireVector we are rendering to a waveform
         :param n: An integer from 0 to segment_len-1
         :param current_val: the value to be rendered
         :param symbol_len: and integer for how big to draw the current value
+        :param repr_func: function to use for representing the current_val;
+            examples are 'hex', 'oct', 'bin', and 'str' (for decimal). Defaults to 'hex'.
 
         Returns a string of printed length symbol_len that will draw the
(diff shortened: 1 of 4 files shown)


timsherwood:
All the backwards compatible changes are fine with me.   A couple thoughts:

* I like the ability to see in other encodings but the "0x" or "0b" at the start of every signal is a lot of wasted space.  Could you add something "(0x)" to the end of the name of the signal to denote its representation instead?   I don't feel to strongly about it though, just something that kind of hopped out to me when looking at the traces

* That is really cool to figure out the length automatically!  I like that a lot, but two things give me a little bit of pause for new people.  a) it is weird to me to have different lengths of symbols across different runs of the program based on the data provided (it is hard enough to get people to understand the default of 5 cycles per chunk in pyrtl) and b) I worry about really long values (and maybe really short ones too).

mdko:
> All the backwards compatible changes are fine with me. A couple thoughts:
> 
> * I like the ability to see in other encodings but the "0x" or "0b" at the start of every signal is a lot of wasted space.  Could you add something "(0x)" to the end of the name of the signal to denote its representation instead?   I don't feel to strongly about it though, just something that kind of hopped out to me when looking at the traces

I think this can be addressed by the user by passing a `repr_per_name` map, like `{w: lambda v: hex(v)[2:] for w in names}` if they really want to get rid of the leading `0x` or `0b` on all the wires.

> * That is really cool to figure out the length automatically!  I like that a lot, but two things give me a little bit of pause for new people.  a) it is weird to me to have different lengths of symbols across different runs of the program based on the data provided (it is hard enough to get people to understand the default of 5 cycles per chunk in pyrtl) and b) I worry about really long values (and maybe really short ones too).

We'll leave `symbol_len=5` the default to behave as before/not make it harder to understand. It will be different lengths only if the user explicitly sets `symbol_len=None`.

PR #393, opened 2021-06-16
Convenience function for importing Verilog into PyRTL via the Yosys/BLIF workflow

mdko:
Rather than always having to remember the correct incantations to Yosys to convert a Verilog file into a PyRTL-importable BLIF file, this PR adds a function that does the Yosys script call for you, then calls `input_from_blif()` like normal.


mdko:
We keep the same interface as `input_from_blif()` for consistency (i.e. taking in an open file descriptor or string containing code). This currently is skipped in CI (because the check for the presence of `yosys` fails); we may want to visit that so that CI installed `yosys`. This will have the happy effect of increasing our code coverage for this newly introduced function.

PR #394, opened 2021-06-17
Fix html trace visualization when value is a boolean and/or supplied …

mdko:
When the wire was 1-bit wide and the value provided was a boolean (True/False) rather than an int, the html trace visualization printed 'True'/'False' instead of '0'/'1'. This PR fixes that.

It also adds a small change to step_multiple() so that int subclasses aren't always cast to ints, thus preserving their subclass information (e.g. IntEnums). This may be helpful in the future for automatically detecting and displaying int subclasses specially in the trace (rather than needing to supply a map in `repr_per_name`).

Now:
Screen Shot 2021-06-16 at 6 58 18 PM

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -7,6 +7,7 @@ import re
 import numbers
 import collections
 import copy
+import six
 
 from .pyrtlexceptions import PyrtlError, PyrtlInternalError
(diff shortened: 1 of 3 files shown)


mdko:
This was before:


Before:
Screen Shot 2021-06-16 at 8 10 17 PM

PR #409, opened 2022-01-29
Remove Python 2 support

mdko:
**Note as of 22-01-29: don't pull yet**

This removes Python 2 support, in preparation for whenever the macOS 12.3 update arrives and we can finally happily drop Python 2. A future commit that naturally follows this could be cleaning up existing code to use any new Python 3.X features.

This is currently set to run tests/work on Python 3.8. We should probably change this to reflect whatever minor version of Python 3 macOS ends up setting as the default (either installed with the system or by included with Xcode).

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -32,5 +32,4 @@ omit =
     .tox/*
     */tests/*
     */ctypes/*
-    six.py
     pyparsing.py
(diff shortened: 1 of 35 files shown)


fdxmw:
I think this should be done as of [`9d1ac75`](https://github.com/UCSBarchlab/PyRTL/commit/9d1ac754b9fb171f9673d3bed12179980efef6e1). Let me know if we're missing anything!

PR #412, opened 2022-01-30
Import ABC from collections.abc for Python 3.10 compatibility.

### LINKED ISSUE #410: Import `collections.abc.Mapping`
To support newer Pythons (>= 3.9), this line:
https://github.com/UCSBarchlab/PyRTL/blob/5de365219705cdc1116f3d7efe938f49e888d449/pyrtl/simulation.py#L1031

Should use `collections.abc.Mapping` instead of `collections.Mapping`.

If you need to support both Python 2.x and Python 3.x, you can use one of the methods proposed here:
https://stackoverflow.com/questions/53978542/how-to-use-collections-abc-from-both-python-3-8-and-python-2-7

tirkarthi:
Fixes #410

### DIFF
--- a/pyrtl/compilesim.py
+++ b/pyrtl/compilesim.py
@@ -4,7 +4,6 @@ import ctypes
 import subprocess
 import tempfile
 import shutil
-import collections
 from os import path
 import platform
 import sys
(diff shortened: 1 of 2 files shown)


mdko (approved):
Great, thanks!

Issue #411, opened 2022-01-29
Missing SDFF_PP1 circuit model yosys primitive


JulianKemmerer:
Hello there

I am using PyRTL with `.blif` files output from `yosys`

Circuits look like so in terms of cell usage:
```
Number of cells:                147
     $_ANDNOT_                      25
     $_AND_                          1
     $_DFF_P_                       69
     $_MUX_                          3
     $_NAND_                         1
     $_NOT_                          1
     $_ORNOT_                       21
     $_OR_                           9
     $_SDFF_PP1_                     1
     $_XNOR_                         7
     $_XOR_                          9
```

For whatever reason yosys uses a ` $_SDFF_PP1_` flip flop of some kind.

And from PyRTL I get
```
  File "/home/julian/.local/lib/python3.8/site-packages/pyrtl/importexport.py", line 397, in extract_model_reference
    subckt = Subcircuit(models[command['model_name']], clk_set=formal_clks, block=block)
KeyError: '$_SDFF_PP1_'
```

Any suggestions on what to do about this `SDFF_PP1`?

Thanks!

timsherwood:
Hi Julian, this is interesting -- Do you have a small example .blif file we could use to test?   I am guessing this is a problem with "reset to 1" style dffs.  Does the source verilog have a dff with that behavior by chance?  PyRTL supports that, but I am not sure we ever revisited the blif converter to take such designs.  That is just a guess though, @mdko has run more designs through the .blif path than anyone else.

mdko:
Here's a simple file that will produce a similar problem (reset to 0, so this produces `$_SDFF_PP0_`'s):
```
module counter (clk, rst, en, count);

    input clk, rst, en;
    output reg [3:0] count;

    always @(posedge clk)
        if (rst)
            count <= 4'd0;
        else if (en)
            count <= count + 4'd1;

endmodule
```

The issue is that the `input_from_verilog()` function calls out to `yosys` to produce the BLIF file, which in turn is just imported using `input_from_blif()`. `input_from_blif()` was previously the only way to do this, and `input_from_verilog()` was added as a wrapper to make developers' lives a little easier by supplying the necessary yosys arguments/commands to produce an importable BLIF (though obviously not in this case).

I'm still trying to figure out why this is happening; I'm not sure if it's because of the recent yosys version updates or something else, but I don't remember this failing previously.

## Solution 1
[...]

PR #413, opened 2022-02-12
Improve BLIF importer to handle more types of dffs

mdko:
This is a fix for #411 (an issue with BLIF import where flip flops are involved).

One of the solutions proposed in #411 was to handle more of the flip-flop formats that Yosys produces (see all of them [here](https://github.com/YosysHQ/yosys/blob/master/techlibs/common/simcells.v)). These changes *don't* handle the forms that use negedges (since PyRTL abstracts that away, we're going to assume posedge anyway), nor the `_DLATCH_PN0_*` latches found near the end of that file.

As future work, more tests should be added that import Verilog exercises all of these types.

### DIFF
--- a/pyrtl/importexport.py
+++ b/pyrtl/importexport.py
@@ -14,11 +14,13 @@ import os
 import subprocess
 import six
 import sys
+import functools
+import operator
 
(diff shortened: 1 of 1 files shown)


mdko:
Closes #411.

PR #416, opened 2022-05-18
Support newer Python graphviz module

ryoon:
Python graphviz 0.19 or later has _repr_image_svg_xml() instead of _repr_svg_().
Fix runtime error from output_to_svg() and block_to_svg().


mdko:
Great, thanks!

Issue #420, opened 2022-06-13
"toplevel" in verilog output is hardcoded


bjourne:
When outputting your design to verilog using `output_to_verilog` the name of the module is always set to "toplevel". It would be great if you could pass the module name as a parameter to that function.

PR #427, opened 2022-09-27
Added Module name change functionality for Issue #420

spencercheese:


fdxmw:
I believe this was implemented in #474.

PR #453, opened 2024-07-02
parameter fix in the documentation for synthesize method (merge_io_vectors)

VarunRao21:
In passes.py, the param documentation says a method is called 'merge_io_wirevectors', while in the code it is called 'merge_io_vectors' (https://github.com/UCSBarchlab/PyRTL/blob/development/pyrtl/passes.py#L468C17-L468C37). This is reflected on the PyRTL docs: https://pyrtl.readthedocs.io/en/latest/analysis.html#synthesis.

### DIFF
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -465,7 +465,7 @@ def synthesize(update_working_block=True, merge_io_vectors=True, block=None):
 
     :param bool update_working_block: Boolean specifying if working block
         should be set to the newly synthesized block.
-    :param bool merge_io_wirevectors: if False, turn all N-bit IO wirevectors
+    :param bool merge_io_vectors: if False, turn all N-bit IO wirevectors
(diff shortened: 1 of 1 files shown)


fdxmw (commented):
Thanks for fixing this! Before I merge your change, please add some more details to the commit message - the current commit message is just "documentation fix", which isn't the most helpful for someone digging through commit history. For example, maybe mention that you're correcting a parameter name in the documentation for `synthesize`?

VarunRao21:
Updated commit message. Thanks for the feedback!

fdxmw:
> Updated commit message. Thanks for the feedback!

Hmm, I still see "documentation fix" as the commit message here https://github.com/UCSBarchlab/PyRTL/pull/453/commits

VarunRao21:
Oops, please try again now.

PR #456, opened 2024-12-11
Adding one-hot to binary helper function

### LINKED ISSUE #455: Add a helper function to encode 1-hot to binary
https://github.com/intel/rohd-hcl/blob/main/doc/components/onehot.md

Implement a 1-hot to binary that takes a wirevector with only 1 bit set to 1 and returns the bit position of that 1 in binary.

vaaniarora:
Takes a one-hot input and returns the bit position of the high bit in binary.

### DIFF
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -13,7 +13,7 @@ from typing import Union, NamedTuple
 from .core import working_block, _NameIndexer, _get_debug_mode, Block
 from .pyrtlexceptions import PyrtlError, PyrtlInternalError
 from .wire import WireVector, Input, Output, Const, Register, WrappedWireVector
-from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list, mux, select
+from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list, select
 
 # -----------------------------------------------------------------
 #        ___       __   ___  __   __
@@ -1684,26 +1684,33 @@ def wire_matrix(component_schema, size: int):
 
     return _WireMatrix
 
+
 def one_hot_to_binary(w) -> WireVector:
     '''Takes a one-hot input and returns the bit position of the high bit in binary.
 
     :param w: WireVector or a WireVector-like object or something that can be converted
-        into a Const (in accordance with the as_wires() required input).
+        into a Const (in accordance with the :py:func:`as_wires()` required input). Example
+        inputs: 0b0010, 64, 0b01.
     :return: The bit position of the high bit in binary as a WireVector.
 
     If the input contains multiple 1s, the bit position of the first 1 will
-        be returned. If the input contains no 1s, 0 will be returned. 
+        be returned. If the input contains no 1s, 0 will be returned.
+
+    Examples::
+
+        one_hot_to_binary(0b0010) # returns 1
+        one_hot_to_binary(64) # returns 6
+        one_hot_to_binary(0b1100) # returns 2, the bit position of the first 1
+        one_hot_to_binary(0) # returns 0
     '''
 
-    if not isinstance(w, WireVector):
-        w = as_wires(w)
+    w = as_wires(w)
 
-    n = len(w)
-    pos = 0
-    found = as_wires(0)
+    pos = 0  # Bit position of the first 1
+    already_found = as_wires(False)  # True if first 1 already found, False otherwise
 
-    for i in range(n):
-        pos = select(found | ~w[i], pos, i)
-        found = select(found, found, w[i])
+    for i in range(len(w)):
+        pos = select(w[i] & ~already_found, i, pos)
+        already_found = already_found | w[i]
 
     return pos
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -13,7 +13,7 @@ from typing import Union, NamedTuple
 from .core import working_block, _NameIndexer, _get_debug_mode, Block
 from .pyrtlexceptions import PyrtlError, PyrtlInternalError
 from .wire import WireVector, Input, Output, Const, Register, WrappedWireVector
-from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list
+from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list, mux, select
 
 # -----------------------------------------------------------------
 #        ___       __   ___  __   __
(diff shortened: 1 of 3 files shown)


pyrtl/helperfuncs.py
```diff
@@ -13,7 +13,7 @@
 from .core import working_block, _NameIndexer, _get_debug_mode, Block
 from .pyrtlexceptions import PyrtlError, PyrtlInternalError
 from .wire import WireVector, Input, Output, Const, Register, WrappedWireVector
-from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list
+from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list, mux, select
```
fdxmw: Is this `mux` import needed?

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
             return len(self._components)
 
     return _WireMatrix
+
+def one_hot_to_binary(w) -> WireVector:
+    '''Takes a one-hot input and returns the bit position of the high bit in binary.
+
+    :param w: WireVector or a WireVector-like object or something that can be converted
+        into a Const (in accordance with the as_wires() required input).
```
fdxmw: I think you can make this reference to `as_wires` a clickable link with :py:func:`as_wires`

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+    If the input contains multiple 1s, the bit position of the first 1 will
+        be returned. If the input contains no 1s, 0 will be returned. 
+    '''
+
+    if not isinstance(w, WireVector):
+        w = as_wires(w)
+
+    n = len(w)
+    pos = 0
+    found = as_wires(0)
```
fdxmw: I think it's a little clearer to initialize to `as_wires(False)` to clarify that `found` is a boolean rather than an integer

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+    :return: The bit position of the high bit in binary as a WireVector.
+
+    If the input contains multiple 1s, the bit position of the first 1 will
+        be returned. If the input contains no 1s, 0 will be returned. 
+    '''
+
+    if not isinstance(w, WireVector):
+        w = as_wires(w)
+
+    n = len(w)
```
fdxmw: `n` is only used once so it seems clearer to inline it below, like `for i in range(len(w)):` This way the reader doesn't have to remember what `n` is :)

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+    pos = 0
```
fdxmw: I think it'd help to add some comments on what `pos` and `found` represent. `found` seems particularly tricky since it seems to indicate whether the `1` was found on a previous iteration through the loop. Maybe it should be called `already_found` or something similar?

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+    found = as_wires(0)
+
+    for i in range(n):
+        pos = select(found | ~w[i], pos, i)
+        found = select(found, found, w[i])
```
fdxmw: I think this is equivalent to `found = found | w[i]` ?

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+
+    if not isinstance(w, WireVector):
+        w = as_wires(w)
+
+    n = len(w)
+    pos = 0
+    found = as_wires(0)
+
+    for i in range(n):
+        pos = select(found | ~w[i], pos, i)
```
fdxmw: I think this is a little easier to read as `select(w[i] & ~found, i, pos)` because it describes the condition where we've found the `1`, rather than the condition where we haven't

tests/test_helperfuncs.py
```diff
@@ -1771,6 +1770,57 @@ def test_byte_matrix_input_concatenate(self):
         self.assertEqual(sim.inspect('byte_matrix[0].high'), 0xA)
         self.assertEqual(sim.inspect('byte_matrix[0].low'), 0xB)
 
+class TestOneHotToBinary(unittest.TestCase):
+    
+    def test_simple_onehot(self):
+        
+        pyrtl.reset_working_block()
+        
+        o1 = pyrtl.WireVector(name='o1')
```
fdxmw: For these kinds of tests, it's a little easier to define your test logic once with placeholder `pyrtl.Input` as your test's input and `pyrtl.Output` as your test's output. Then you can set the test inputs with each call to `sim.step()`, with its `provided_inputs` argument. This moves each test case's inputs closer to its outputs in the test file, which makes the test a little easier to read. Same comment for the other new tests below :)

tests/test_helperfuncs.py
```diff
@@ -1771,6 +1770,57 @@ def test_byte_matrix_input_concatenate(self):
         self.assertEqual(sim.inspect('byte_matrix[0].high'), 0xA)
         self.assertEqual(sim.inspect('byte_matrix[0].low'), 0xB)
 
+class TestOneHotToBinary(unittest.TestCase):
+    
+    def test_simple_onehot(self):
+
```
fdxmw: Remove these blank lines after each test definition, for consistency with the existing tests in this file. I think the blank line after the class definition should also be deleted (lines 1774, 1776)

tests/test_helperfuncs.py
```diff
@@ -1771,6 +1770,57 @@ def test_byte_matrix_input_concatenate(self):
         self.assertEqual(sim.inspect('byte_matrix[0].high'), 0xA)
         self.assertEqual(sim.inspect('byte_matrix[0].low'), 0xB)
 
+class TestOneHotToBinary(unittest.TestCase):
+    
+    def test_simple_onehot(self):
+        
+        pyrtl.reset_working_block()
```
fdxmw, line 1777: You can move these `reset_working_block()` calls to a shared `setUp()` method, there are some examples in this file

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
             return len(self._components)
 
     return _WireMatrix
+
+def one_hot_to_binary(w) -> WireVector:
+    '''Takes a one-hot input and returns the bit position of the high bit in binary.
```
fdxmw, line 1689: Consider adding some example inputs and outputs to the documentation

pyrtl/helperfuncs.py
```diff
@@ -1683,3 +1683,27 @@ def __len__(self):
...
+
+    :param w: WireVector or a WireVector-like object or something that can be converted
+        into a Const (in accordance with the as_wires() required input).
+    :return: The bit position of the high bit in binary as a WireVector.
+
+    If the input contains multiple 1s, the bit position of the first 1 will
+        be returned. If the input contains no 1s, 0 will be returned. 
+    '''
+
+    if not isinstance(w, WireVector):
```
fdxmw: I don't think this check is needed since `as_wires` should pass through `w` if it's already a `WireVector`

PR #458, opened 2025-01-03
Improve compilesim.py test coverage

gaborszita:
I added more tests to improve the test coverage of `compilesim.py`:

- Added tests `test_less_than_cmp_simulation` and `test_equals_simulation` to cover functions `_build_eq` and `_build_cmp` in `compilesim.py`
- Added tests to test RomBlocks with more bit widths to cover the function `_romwidth` in `compilesim.py`
- Added test `test_rom_romblock_in_memory_value_map_error` to test the exception `raise PyrtlError('RomBlock in memory_value_map')` (line 781 in `compilesim.py`)

### DIFF
--- a/tests/test_compilesim.py
+++ b/tests/test_compilesim.py
@@ -990,16 +990,16 @@ class RomBlockSimBase(unittest.TestCase):
                                                  ("o2", lambda x: rom_data_function(2 * x))), 6)
         self.compareIO(self.sim_trace, exp_out)
 
-    def test_function_rom_block_with_optimization_bidwidth_4(self):
+    def test_function_rom_block_with_optimization_bitwidth_4(self):
         self.rom_block_with_optimization_helper(4)
 
-    def test_function_rom_block_with_optimization_bidwidth_12(self):
+    def test_function_rom_block_with_optimization_bitwidth_12(self):
         self.rom_block_with_optimization_helper(12)
 
-    def test_function_rom_block_with_optimization_bidwidth_32(self):
+    def test_function_rom_block_with_optimization_bitwidth_32(self):
         self.rom_block_with_optimization_helper(32)
 
-    def test_function_rom_block_with_optimization_bidwidth_64(self):
+    def test_function_rom_block_with_optimization_bitwidth_64(self):
         self.rom_block_with_optimization_helper(64)
 
     def test_rom_out_of_range_error(self):
--- a/tests/test_compilesim.py
+++ b/tests/test_compilesim.py
@@ -914,13 +914,13 @@ class RomBlockSimBase(unittest.TestCase):
         sim_trace_a.print_trace(output, compact=True)
         self.assertEqual(output.getvalue(), expected_output)
 
-    def romBlock_test_helper(self, bidwidth):
+    def rom_block_test_helper(self, bitwidth):
 
         def rom_data_function(add):
             return int((add + 5) / 2)
 
         pyrtl.reset_working_block()
-        self.bitwidth = bidwidth
+        self.bitwidth = bitwidth
         self.addrwidth = 4
         self.output1 = pyrtl.Output(self.bitwidth, "o1")
         self.output2 = pyrtl.Output(self.bitwidth, "o2")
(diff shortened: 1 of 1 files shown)


tests/test_compilesim.py
```diff
@@ -902,13 +914,13 @@ def compareIO(self, sim_trace_a, expected_output):
         sim_trace_a.print_trace(output, compact=True)
         self.assertEqual(output.getvalue(), expected_output)
 
-    def test_function_RomBlock(self):
+    def romBlock_test_helper(self, bidwidth):
```
fdxmw: I know you didn't start this, but we should try to avoid mixing capitalization styles. So this should be called `rom_block_test_helper`. There are a few more `romBlock`s in this change that should be `rom_block`s. Also there's a typo, `bidwith` should be `bitwidth` (happens again a few more times below)

tests/test_compilesim.py
```diff
@@ -977,6 +1013,19 @@ def test_rom_out_of_range_error(self):
         with self.assertRaises(pyrtl.PyrtlError) as ex:
             self.sim(tracer=sim_trace)
 
+    def test_rom_romblock_in_memory_value_map_error(self):
+        rom_data_array = [1, 2, 3, 4, 5, 6, 7, 8]
```
fdxmw: I don't think the value of `rom_data_array` or `mem_val_map` matter for this test? If that's right, consider simplifying them (reduce addrwidth to 1?), so it's more obvious to the reader that these values are not important

tests/test_compilesim.py
```diff
@@ -977,6 +1013,19 @@ def test_rom_out_of_range_error(self):
...
+        rom1 = pyrtl.RomBlock(bitwidth=4, addrwidth=3, romdata=rom_data_array)
+        rom_add = pyrtl.Input(3, "rom_in_2")
```
fdxmw: Call this `rom_addr` to be consistent with `addrwidth` ? Also it seems better to have the wire name match the variable name, so
```
rom_addr = pyrt.Input(3, "rom_addr")
```
Same comment for `rom_out_2` below

tests/test_compilesim.py
```diff
@@ -977,6 +1013,19 @@ def test_rom_out_of_range_error(self):
...
+        rom_out = pyrtl.Output(4, "rom_out_2")
+        rom_out <<= rom1[rom_add]
+        mem_val_map = {rom1: {0: 0, 1: 1, 2: 2, 3: 3}}
+
+        sim_trace = pyrtl.SimulationTrace()
+        with self.assertRaises(pyrtl.PyrtlError) as error:
+            self.sim(tracer=sim_trace, memory_value_map=mem_val_map)
+        self.assertEqual(str(error.exception), "RomBlock in memory_value_map")
```
fdxmw: Avoid checking for exact error messages in tests because it tightly couples the error code and test code - a minor spelling/grammar change to the error message will break this test, which makes it more difficult to modify PyRTL. I think the `assertRaises` is enough of a test, but if you want to be more specific maybe just check that the error message mentions `memory_value_map` somewhere?

fdxmw (commented):
Thank you for the pull request, looks good overall! I have some minor style comments, and there are a couple `pycodestyle` errors in the presubmit checks too:

```
  tests/test_compilesim.py:948:1: W293 blank line contains whitespace
  tests/test_compilesim.py:951:1: W293 blank line contains whitespace
```

gaborszita:
@fdxmw fixed

tests/test_compilesim.py
```diff
@@ -966,6 +990,18 @@ def rom_data_function(add):
                                                  ("o2", lambda x: rom_data_function(2 * x))), 6)
         self.compareIO(self.sim_trace, exp_out)
 
+    def test_function_rom_block_with_optimization_bidwidth_4(self):
```
fdxmw: Sorry but there are still `bidwidth`s in the names of these four tests

gaborszita: What should I name the tests? I added tests with different bitwidths to cover the `_romwidth` function in `compilesim.py`.

gaborszita: Oh a `bidwith` typo, sorry I didn't understand what you mean, I'll fix it

fdxmw (approved):

Issue #460, opened 2025-01-11
Potential unneeded exception


gaborszita:
I was writing some tests for `memory.py` and I think that one of the checks that throws an exception is unnecessary.

In the `_assignment` method in `memory.py`, we have this check that throws an exception:

```
item = as_wires(item, bitwidth=self.addrwidth, truncating=False)
if len(item) > self.addrwidth: 
    raise PyrtlError('error, the wire indexing the memory bitwidth > addrwidth')
```

I tried to reproduce this exception, but I wasn't able to. So, looked at the code and realized that this exception can't happen. This is because `len(item) > self.addrwidth` is always checked earlier in the `__getitem__` method and an exception is thrown if it is true. My finding is explained below in more detail below.

The only place where the `_assignment` method is called is in the `__setitem__` method:

```
if isinstance(assignment, _MemAssignment):
    self._assignment(item, assignment.rhs, is_conditional=assignment.is_conditional)
```

We can see here that the `_assignment` method can only get called if `assignment` is an instance of `_MemAssignment`.

Furthermore, the only place where a `_MemAssignment` object is instantiated is in the `_MemIndexed` class:

```
def __ilshift__(self, other):
        return _MemAssignment(rhs=other, is_conditional=False)

def __ior__(self, other):
        return _MemAssignment(rhs=other, is_conditional=True)
```

Meanwhile, the only place where a `_MemIndexed` object is instantiated and returned to the user is in the `__getitem__` method:

```
def __getitem__(self, item) -> WireVector:
    """Create a read port to load items from the MemBlock."""
    item = as_wires(item, bitwidth=self.addrwidth, truncating=False)
    if len(item) > self.addrwidth:
        raise PyrtlError('memory index bitwidth > addrwidth')
    return _MemIndexed(mem=self, index=item)
```

As we can see, `len(item) > self.addrwidth` is already checked in this method and an exception is returned if it is true.

So, I think this check in the `_assignment` method is unnecessary:

```
item = as_wires(item, bitwidth=self.addrwidth, truncating=False)
if len(item) > self.addrwidth: 
    raise PyrtlError('error, the wire indexing the memory bitwidth > addrwidth')
```

Or am I missing something? I spent a lot of time trying to reproduce this exception, but I wasn't able to.

fdxmw:
I think you're right! It looks like we can also remove the call to `as_wires` in `_assignment`, since that already happens in `__getitem__`? 

[...]

PR #461, opened 2025-01-11
Add more tests for MemBlock and _MemIndexed

gaborszita:
Classes `MemBlock` and `_MemIndexed` are now fully covered by tests, except one exception that I don't think can happen (see #460)

### DIFF
--- a/tests/test_memblock.py
+++ b/tests/test_memblock.py
@@ -274,12 +274,6 @@ class MemIndexedTests(unittest.TestCase):
         self.mem_2 = pyrtl.MemBlock(bitwidth=16, addrwidth=1)
         self.assertEqual(len(self.mem_2[0]), 16)
 
-    def reassemble_bits(bits: list[int]):
-        value = 0
-        for bit in bits:
-            value = (value << 1) | bit
-        return value
-
     def test_memindexed_getitem(self):
         mem = pyrtl.MemBlock(bitwidth=8, addrwidth=1, max_read_ports=None)
         mem_addr = pyrtl.Input(1, 'mem_addr')
--- a/tests/test_memblock.py
+++ b/tests/test_memblock.py
@@ -101,6 +101,44 @@ class RTLMemBlockDesignBase(unittest.TestCase):
         mem = pyrtl.MemBlock(32, 8)
         self.assertIs(pyrtl.working_block().get_memblock_by_name(mem.name), mem)
 
+class RTLMemBlockErrorTests(unittest.TestCase):
+    def setUp(self):
+        pyrtl.reset_working_block()
+
+    def test_negative_bitwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(-1, 1)
+
+    def test_negative_addrwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(1, -1)
+    
+    def test_memindex_bitwidth_more_than_addrwidth(self):
+        mem = pyrtl.MemBlock(1, 1)
+        mem_in = pyrtl.Input(2, 'mem_in')
+        mem_out = pyrtl.Output(1, 'mem_out')
+        with self.assertRaises(pyrtl.PyrtlError):
+            mem_out <<= mem[mem_in]
+
+    def test_memblock_write_data_larger_than_memory_bidwidth(self):
+        mem = pyrtl.MemBlock(1, 1)
+        mem_addr = pyrtl.Input(1, 'mem_addr')
+        mem_in = pyrtl.Input(2, 'mem_in')
+        with self.assertRaises(pyrtl.PyrtlError):
+            mem[mem_addr] <<= mem_in
+
+    def test_memblock_enable_signal_not_1_bit(self):
+        mem = pyrtl.MemBlock(1, 1)
+        mem_addr = pyrtl.Input(1, 'mem_addr')
+        mem_in = pyrtl.Input(1, 'mem_in')
+        with self.assertRaises(pyrtl.PyrtlError):
+            mem[mem_addr] <<= pyrtl.MemBlock.EnabledWrite(mem_in, enable=pyrtl.Input(2))
+
+    def test_read_ports_exception(self):
+        mem = pyrtl.MemBlock(1, 1)
+        with self.assertRaises(pyrtl.PyrtlError):
+            mem.read_ports()
+
 
 class MemIndexedTests(unittest.TestCase):
     def setUp(self):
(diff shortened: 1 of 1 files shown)


tests/test_memblock.py
```diff
@@ -102,6 +102,45 @@ def test_memblock_added_default_named(self):
         self.assertIs(pyrtl.working_block().get_memblock_by_name(mem.name), mem)
 
 
+class RTLMemBlockErrorTests(unittest.TestCase):
+    def setUp(self):
+        pyrtl.reset_working_block()
+
+    def test_negative_bitwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(-1, 1)
```
fdxmw: Lines like these are a little easier to read with keyword arguments (`pyrtl.MemBlock(bitwidth=-1, addrwidth=1)`) as the reader may not remember whether `bitwidth` or `addrwidth` comes first. 

If you agree and want to change this, please change it consistently throughout this commit

tests/test_memblock.py
```diff
@@ -102,6 +102,45 @@ def test_memblock_added_default_named(self):
...
+            pyrtl.MemBlock(1, -1)
+
+    def test_memindex_bitwidth_more_than_addrwidth(self):
+        mem = pyrtl.MemBlock(1, 1)
+        mem_in = pyrtl.Input(2, 'mem_in')
+        mem_out = pyrtl.Output(1, 'mem_out')
+        with self.assertRaises(pyrtl.PyrtlError):
+            mem_out <<= mem[mem_in]
+
+    def test_memblock_write_data_larger_than_memory_bidwidth(self):
```
fdxmw: typo: `bidwidth` :)

tests/test_memblock.py
```diff
@@ -102,6 +102,45 @@ def test_memblock_added_default_named(self):
...
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(-1, 1)
+
+    def test_negative_addrwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(1, -1)
+
+    def test_memindex_bitwidth_more_than_addrwidth(self):
+        mem = pyrtl.MemBlock(1, 1)
+        mem_in = pyrtl.Input(2, 'mem_in')
```
fdxmw: The name `mem_in` is confusing in this commit as it's sometimes the address, and sometimes the data to write. Seems clearer to always call the address `mem_addr`?

tests/test_memblock.py
```diff
@@ -102,6 +102,45 @@ def test_memblock_added_default_named(self):
...
+
+    def test_negative_bitwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(-1, 1)
+
+    def test_negative_addrwidth(self):
+        with self.assertRaises(pyrtl.PyrtlError):
+            pyrtl.MemBlock(1, -1)
+
+    def test_memindex_bitwidth_more_than_addrwidth(self):
```
fdxmw: Maybe call this `greater_than` instead of `more_than` ?

Regardless, the word choice here should be consistent with the next test (`more` vs `larger`)

tests/test_memblock.py
```diff
@@ -229,6 +268,53 @@ def test_write_memindexed_ior(self):
...
+        self.assertEqual(len(self.mem_2[0]), 16)
+
+    def test_memindexed_getitem(self):
+        mem = pyrtl.MemBlock(bitwidth=8, addrwidth=1, max_read_ports=None)
+        mem_in = pyrtl.Input(1, 'mem_in')
+        mem_out_array = [pyrtl.Output(8, 'mem_out_' + str(i)) for i in range(8)]
+        for i in range(8):
+            mem_out_array[i] <<= mem[mem_in][i]
+        mem_value_map = {mem: {0: 7, 1: 5}}
+        sim_trace = pyrtl.SimulationTrace()
```
fdxmw: Creating a `SimulationTrace` shouldn't be needed here since `Simulation` should create a `SimulationTrace` for you if none is provided (happens again in the next two tests)

tests/test_memblock.py
```diff
@@ -229,6 +268,53 @@ def test_write_memindexed_ior(self):
...
+        sim = pyrtl.Simulation(tracer=sim_trace, memory_value_map=mem_value_map)
+        for i in range(len(mem_value_map[mem])):
```
fdxmw: Doesn't seem like we gain much by testing both addresses? Consider just checking the first address as loops like these make it harder to debug the test. When a test failure occurs, we won't know the values of `i` or `j`

tests/test_memblock.py
```diff
@@ -229,6 +268,53 @@ def test_write_memindexed_ior(self):
...
+            sim.step({mem_in: i})
+            binary = bin(mem_value_map[mem][i])[2:].zfill(8)
+            for j in range(8):
+                self.assertEqual(sim.inspect(mem_out_array[j]), int(binary[7 - j]))
```
fdxmw: Similarly, it's a little better to reassemble the inspected binary value, and then do one big `assertEqual`, rather than doing one `assertEqual` for each bit in the loop. With one big `assertEqual`, the test failure error message will be better, since we can more easily see which bits are wrong. With the single-bit `assertEqual`, the error message will just tell us that 0 is not equal to 1 or vice versa.

You can reassemble the inspected values with something like:
```
>>> def reassemble_bits(bits: list[int]):
...     value = 0
...     for bit in bits:
...         value = (value << 1) | bit
...     return value
>>> reassemble_bits([0, 0])
0
>>> reassemble_bits([0, 1])
1
>>> reassemble_bits([1, 0])
2
>>> reassemble_bits([1, 1])
3
```

If you want to keep things the way they are, consider moving this logic to check if the Nth bit is set to a helper function, as it's a bit complicated and it's repeated in the next two tests.

Also you can check this a little more easily with bitmasks and bitwise AND, something like
```
>>> def check_nth_bit(value: int, n: int):
...     return bool(value & (1 << n))
>>> for n in range(8):
...     print(n, check_nth_bit(27, n))
...
0 True
1 True
2 False
3 True
4 True
5 False
6 False
7 False
```

fdxmw (commented):
Thank you for improving these tests! I have some minor comments

tests/test_memblock.py
gaborszita: I changed the name to `test_memindex_bitwidth_greater_than_addrwidth`, but I left `test_memblock_write_data_larger_than_memory_bitwidth` as it is because I think when we talk about numbers it is better to say greater and less and when we talk about data it is better to say larger and smaller. But I can change it if you disagree.

fdxmw (approved):

PR #462, opened 2025-01-12
Eliminate double inverts in pyrtl.optimize

### LINKED ISSUE #447: Optimize is currently missing the elimination of "double not"
The pyrtl.optimize code is not eliminating "double not" conditions (reducing them to a single wire).   Right now it does some more complex optimizations including CSE but this very simple case gets missed. You can see very clearly from the code below.

```python
import pyrtl

a = pyrtl.Input(1,'a')
q = pyrtl.Output(1,'q')
q <<= ~(~(a))

pyrtl.synthesize()

with open('pre-optimize.svg','w') as f:
    pyrtl.output_to_svg(f)

pyrtl.optimize()

with open('post-optimize.svg','w') as f:
    pyrtl.output_to_svg(f)
```

gaborszita:
Closes #447

Example:

| Block before optimize | Block after optimize (without my changes) | Block after optimize (with my changes) |
| -------- | ------- | ------- |
| image  | image | image |

### DIFF
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -52,14 +52,25 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
         constant_propagation(block, True)
         _remove_unlistened_nets(block)
         common_subexp_elimination(block)
-        _remove_inverter_chains(block, skip_sanity_check)
+        _optimize_inverter_chains(block, skip_sanity_check)
         if (not skip_sanity_check) or _get_debug_mode():
             block.sanity_check()
     return block
 
 
 def _get_inverter_chains(wire_creator, wire_users):
-    """Returns all inverter chains in the block"""
+    """Returns all inverter chains in the block.
+
+    The function returns a list of inverter chains in the block.
+    Each inverter chain is represented as a list of the LogicNets
+    in the chain.
+
+    Consider the following circuit, for example:
+    A -~-> B -~-> C -w-> X
+    D -~-> E -w-> Y
+    If the function is called on this circuit, it will return
+    [[A, B, C], [D, E]].
+    """
 
     # Build a list of inverter chains. Each inverter chain is a list of WireVectors,
     # from source to destination.
@@ -116,8 +127,8 @@ def _get_inverter_chains(wire_creator, wire_users):
     return inverter_chains
 
 
-def _remove_inverter_chains(block, skip_sanity_check=False):
-    """ Removes all inverter chains from the block.
+def _optimize_inverter_chains(block, skip_sanity_check=False):
+    """ Optimizes inverter chains in the block.
 
     An inverter chain means two or more inverters directly connected
     to each other. Inverter chains are redundant and can be removed.
@@ -155,9 +166,9 @@ def _remove_inverter_chains(block, skip_sanity_check=False):
     # This is the optimized version of the circuit:
     # A -w-> X
     # A -w-> Y
-    # The inverter chains found will be B-C and D-E (two separate chains will be
-    # found instead of B-C-D-E because C has an intermediate user). In the dict,
-    # C will be mapped to A and E will be maped to C. Hence, when finding the
+    # The inverter chains found will be A-B-C and C-D-E (two separate chains will be
+    # found instead of A-B-C-D-E because C has an intermediate user). In the dict,
+    # C will be mapped to A and E will be mapped to C. Hence, when finding the
     # replacement of E, we have to first query the dict to get C, and then query
     # the dict again on C to get A.
     wire_src_dict = _ProducerList()
@@ -179,9 +190,9 @@ def _remove_inverter_chains(block, skip_sanity_check=False):
             # Map the end wire of the inverter chain to the beginning wire.
             wire_src_dict[inverter_chain[-1]] = inverter_chain[start_idx - 1]
 
-    # This loop recreates the LogicNet with inverter chains removed. It adds each
-    # block in the original LogicNet to the new LogicNet if it is not marked for
-    # removal, and replaces the source of the block if its source was the end wire
+    # This loop recreates the block with inverter chains removed. It adds each
+    # LogicNet in the original block to the new block if it is not marked for
+    # removal, and replaces the source of the LocigNet if its source was the end wire
     # of a removed inverter chain.
     for net in block.logic:
         if net not in net_removal_set:
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -61,35 +61,81 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
 def _remove_double_inverts(block, skip_sanity_check=False):
     """ Removes all double invert nets from the block. """
 
-    # checks if the wirevector is used at a LogicNet other than used_at_nets
-    def is_wirevector_used_elsewhere(wire, used_at_nets):
-        for net in block:
-            if net not in used_at_nets:
-                if wire.name in [x.name for x in net.args] \
-                        or wire.name in [x.name for x in net.dests]:
-                    return True
-        return False
+    # wire_creator maps from WireVector to the LogicNet that defines its value.
+    # wire_users maps from WireVector to a list of LogicNets that use its value.
+    wire_creator, wire_users = block.net_connections()
+
+    # Build a list of inverter chains. Each inverter chain is a list of WireVectors,
+    # from source to destination.
+    inverter_wirenet_chains = []
+    for current_dest, current_creator in wire_creator.items():
+        if current_creator.op != "~":
+            # Skip non-inverters.
+            continue
+
+        # The current inverter connects current_arg (a WireVector) to current_dest (also
+        # a WireVector).
+        current_arg = current_creator.args[0]
+        # current_users is the number of LogicNets that use current_dest.
+        current_users = len(wire_users[current_dest])
+
+        # Add the current inverter to the end of this inverter chain.
+        append_to = None
+        # Add the current inverter to the beginning of this inverter chain.
+        prepend_to = None
+        next_inverter_chains = []
+        for inverter_wirenet_chain in inverter_wirenet_chains:
+            chain_arg = inverter_wirenet_chain[0]
+            chain_dest = inverter_wirenet_chain[-1]
+            chain_users = len(wire_users[chain_dest])
+
+            if chain_dest is current_arg and chain_users == 1:
+                # This chain's only destination is the current inverter. Append the
+                # current inverter to the chain.
+                append_to = inverter_wirenet_chain
+            elif chain_arg is current_dest and current_users <= 1:
+                # This chain's only argument is the current inverter. Add the current
+                # inverter to the beginning of the chain.
+                prepend_to = inverter_wirenet_chain
+            else:
+                next_inverter_chains.append(inverter_wirenet_chain)
+
+        #print("current inverter: ", current_arg, "->", current_dest)
+        if append_to and prepend_to:
+            # The current inverter joins two existing inverter chains.
+            next_inverter_chains.append(append_to + prepend_to)
+            #print("  joined", next_inverter_chains)
+        elif append_to:
+            # Add the current inverter after 'append_to'.
+            next_inverter_chains.append(append_to + [current_dest])
+            #print("  appended", next_inverter_chains)
+        elif prepend_to:
+            # Add the current inverter before 'prepend_to'.
+            next_inverter_chains.append([current_arg] + prepend_to)
+            #print("  prepended", next_inverter_chains)
+        else:
+            next_inverter_chains.append([current_arg, current_dest])
+            #print("  start new chain", next_inverter_chains)
+
+        inverter_wirenet_chains = next_inverter_chains
 
     new_logic = set()
     net_exclude_set = set()  # removed nets
     wire_removal_set = set()
-    # Dictionary, key is the destination wire of the invert net, value is the invert net
-    invert_destination_wires = {}
-    for net in block:
-        if net.op == "~":
-            invert_destination_wires[net.dests[0].name] = net
+    for inverter_wirenet_chain in inverter_wirenet_chains:
+        if len(inverter_wirenet_chain) > 1:
+            if len(inverter_wirenet_chain) % 2 == 1: # even number of inverters in a chain
+                end_idx = len(inverter_wirenet_chain) - 1
+            else: # odd number of inverters in a chain
+                end_idx = len(inverter_wirenet_chain) - 2
+            wires_to_remove = inverter_wirenet_chain[1:end_idx]
+            new_logic.add(LogicNet('w', None, args=(inverter_wirenet_chain[0],), \
+                                    dests=(inverter_wirenet_chain[end_idx],)))
+            inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
+            inverters_to_remove.add(wire_creator[inverter_wirenet_chain[end_idx]])
 
-    for net in invert_destination_wires.values():
-        # If the argument of the net is in invert_destination_wires, then it is a double invert
-        # If the net is in net_exclude_set, then it was already removed, so we do not process it
-        if net.args[0].name in invert_destination_wires and net not in net_exclude_set:
-            previous_net = invert_destination_wires[net.args[0].name]
-            if not is_wirevector_used_elsewhere(net.args[0], (net, previous_net)) \
-                    and previous_net not in net_exclude_set:
-                new_logic.add(LogicNet('w', None, args=previous_net.args, dests=net.dests))
-                wire_removal_set.add(net.args[0])
-                net_exclude_set.add(net)
-                net_exclude_set.add(previous_net)
+        wire_removal_set.update(wires_to_remove)
+        net_exclude_set.update(inverters_to_remove)
 
     for net in block.logic:
         if net not in net_exclude_set:
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -136,9 +136,7 @@ def _remove_double_inverts(block, skip_sanity_check=False):
             else:  # odd number of inverters in a chain
                 end_idx = len(inverter_chain) - 2
             wires_to_remove = inverter_chain[1:end_idx+1]
-            for chain_dest in wire_users[inverter_chain[end_idx]]:
-                for arg in chain_dest.args:
-                    wire_src_dict[arg] = inverter_chain[0]
+            wire_src_dict[inverter_chain[end_idx]] = inverter_chain[0]
             inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
             inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
             wire_removal_set.update(wires_to_remove)
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -67,7 +67,7 @@ def _remove_double_inverts(block, skip_sanity_check=False):
 
     # Build a list of inverter chains. Each inverter chain is a list of WireVectors,
     # from source to destination.
-    inverter_wirenet_chains = []
+    inverter_chains = []
     for current_dest, current_creator in wire_creator.items():
         if current_creator.op != "~":
             # Skip non-inverters.
@@ -84,61 +84,55 @@ def _remove_double_inverts(block, skip_sanity_check=False):
         # Add the current inverter to the beginning of this inverter chain.
         prepend_to = None
         next_inverter_chains = []
-        for inverter_wirenet_chain in inverter_wirenet_chains:
-            chain_arg = inverter_wirenet_chain[0]
-            chain_dest = inverter_wirenet_chain[-1]
+        for inverter_chain in inverter_chains:
+            chain_arg = inverter_chain[0]
+            chain_dest = inverter_chain[-1]
             chain_users = len(wire_users[chain_dest])
 
             if chain_dest is current_arg and chain_users == 1:
                 # This chain's only destination is the current inverter. Append the
                 # current inverter to the chain.
-                append_to = inverter_wirenet_chain
+                append_to = inverter_chain
             elif chain_arg is current_dest and current_users <= 1:
                 # This chain's only argument is the current inverter. Add the current
                 # inverter to the beginning of the chain.
-                prepend_to = inverter_wirenet_chain
+                prepend_to = inverter_chain
             else:
-                next_inverter_chains.append(inverter_wirenet_chain)
+                next_inverter_chains.append(inverter_chain)
 
-        #print("current inverter: ", current_arg, "->", current_dest)
         if append_to and prepend_to:
             # The current inverter joins two existing inverter chains.
             next_inverter_chains.append(append_to + prepend_to)
-            #print("  joined", next_inverter_chains)
         elif append_to:
             # Add the current inverter after 'append_to'.
             next_inverter_chains.append(append_to + [current_dest])
-            #print("  appended", next_inverter_chains)
         elif prepend_to:
             # Add the current inverter before 'prepend_to'.
             next_inverter_chains.append([current_arg] + prepend_to)
-            #print("  prepended", next_inverter_chains)
         else:
             next_inverter_chains.append([current_arg, current_dest])
-            #print("  start new chain", next_inverter_chains)
 
-        inverter_wirenet_chains = next_inverter_chains
+        inverter_chains = next_inverter_chains
 
     new_logic = set()
-    net_exclude_set = set()  # removed nets
+    net_removal_set = set()
     wire_removal_set = set()
-    for inverter_wirenet_chain in inverter_wirenet_chains:
-        if len(inverter_wirenet_chain) > 2:
-            if len(inverter_wirenet_chain) % 2 == 1: # even number of inverters in a chain
-                end_idx = len(inverter_wirenet_chain) - 1
-            else: # odd number of inverters in a chain
-                end_idx = len(inverter_wirenet_chain) - 2
-            wires_to_remove = inverter_wirenet_chain[1:end_idx]
-            new_logic.add(LogicNet('w', None, args=(inverter_wirenet_chain[0],), \
-                                    dests=(inverter_wirenet_chain[end_idx],)))
+    for inverter_chain in inverter_chains:
+        if len(inverter_chain) > 2:
+            if len(inverter_chain) % 2 == 1:  # even number of inverters in a chain
+                end_idx = len(inverter_chain) - 1
+            else:  # odd number of inverters in a chain
+                end_idx = len(inverter_chain) - 2
+            wires_to_remove = inverter_chain[1:end_idx]
+            new_logic.add(LogicNet('w', None, args=(inverter_chain[0],),
+                                   dests=(inverter_chain[end_idx],)))
             inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
-            inverters_to_remove.add(wire_creator[inverter_wirenet_chain[end_idx]])
-
+            inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
             wire_removal_set.update(wires_to_remove)
-            net_exclude_set.update(inverters_to_remove)
+            net_removal_set.update(inverters_to_remove)
 
     for net in block.logic:
-        if net not in net_exclude_set:
+        if net not in net_removal_set:
             new_logic.add(net)
 
     block.logic = new_logic
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -140,7 +140,6 @@ def _remove_double_inverts(block, skip_sanity_check=False):
             wire_removal_set.update(wires_to_remove)
             # remove inverters used in the chain
             inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
-            inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
             net_removal_set.update(inverters_to_remove)
             # map the end wire of the inverter chain to the beginning wire
             wire_src_dict[inverter_chain[end_idx]] = inverter_chain[0]
--- a/pyrtl/passes.py
+++ b/pyrtl/passes.py
@@ -58,12 +58,8 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
     return block
 
 
-def _remove_double_inverts(block, skip_sanity_check=False):
-    """ Removes all double invert nets from the block. """
-
-    # wire_creator maps from WireVector to the LogicNet that defines its value.
-    # wire_users maps from WireVector to a list of LogicNets that use its value.
-    wire_creator, wire_users = block.net_connections()
+def _get_inverter_chains(wire_creator, wire_users):
+    """Returns all inverter chains in the block"""
 
     # Build a list of inverter chains. Each inverter chain is a list of WireVectors,
     # from source to destination.
@@ -93,11 +89,13 @@ def _remove_double_inverts(block, skip_sanity_check=False):
                 # This chain's only destination is the current inverter. Append the
                 # current inverter to the chain.
                 append_to = inverter_chain
-            elif chain_arg is current_dest and current_users <= 1:
+            elif chain_arg is current_dest and current_users == 1:
                 # This chain's only argument is the current inverter. Add the current
                 # inverter to the beginning of the chain.
                 prepend_to = inverter_chain
             else:
+                # The current inverter is not connected to the inverter chain, so we
+                # pass the inverter chain through to next_inverter_chains
                 next_inverter_chains.append(inverter_chain)
 
         if append_to and prepend_to:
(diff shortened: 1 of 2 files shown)


fdxmw (commented):
Thank you for fixing this issue! A couple high-level comments before we get into coding details:

1. O(N^2) nested `for` loops over `block.logic` should generally be avoided in passes because they don't scale well as the number of nets increases - the pass will take too long to run. It seems possible to iterate over `block.logic` once and build a list of all the inverter chains in the block (see how other passes create a `_ProducerList` for example). Then I think we can iterate over that list and decide if each inverter chain can be simplified?
1. This really needs a test :) Some test cases that come to mind: 
   1. Chains of 1, 2, 3, and 4 inverters (even numbers of inverters should be removed, and odd numbers of inverters should be simplified to one inverter)
   2. A chain of 2 inverters with another user of the first inverter's output (should not be simplified)

gaborszita:
> Thank you for fixing this issue! A couple high-level comments before we get into coding details:
> 
> 1. O(N^2) nested `for` loops over `block.logic` should generally be avoided in passes because they don't scale well as the number of nets increases - the pass will take too long to run. It seems possible to iterate over `block.logic` once and build a list of all the inverter chains in the block (see how other passes create a `_ProducerList` for example). Then I think we can iterate over that list and decide if each inverter chain can be simplified?
> 2. This really needs a test :) Some test cases that come to mind:
>    
>    1. Chains of 1, 2, 3, and 4 inverters (even numbers of inverters should be removed, and odd numbers of inverters should be simplified to one inverter)
>    2. A chain of 2 inverters with another user of the first inverter's output (should not be simplified)

All addressed

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+
+    new_logic = set()
+    net_exclude_set = set()  # removed nets
+    wire_removal_set = set()
+    # Dictionary, key is the destination wire of the invert net, value is the invert net
+    invert_destination_wires = {}
+    for net in block.logic:
+        if net.op == "~":
+            invert_destination_wires[net.dests[0].name] = net
+    # If double invert nets are removed randomly, this may leave some double inverts behind.
```
fdxmw: We iterate over the block in topological order (see comment here https://github.com/UCSBarchlab/PyRTL/blob/5b282f367baffa2e434960652419f2b4d856d405/pyrtl/core.py#L595) so the inverters will be added to `invert_destination_wires` in topological order. So this code shouldn't randomly remove inverters in this loop, since dictionaries preserve insertion order, see the bottom of this section https://docs.python.org/3/library/stdtypes.html#typesmapping

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+        return False
+
+    new_logic = set()
+    net_exclude_set = set()  # removed nets
+    wire_removal_set = set()
+    # Dictionary, key is the destination wire of the invert net, value is the invert net
+    invert_destination_wires = {}
+    for net in block.logic:
+        if net.op == "~":
+            invert_destination_wires[net.dests[0].name] = net
```
fdxmw: I think we can simplify this by mapping from the destination wirevector itself, rather than its name? `invert_destination_wires[net.dests[0]] = net`

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    # the invert_destination_wires dictionary multiple times, hence the outer while loop.
+    repeat = True
+    while repeat:
+        repeat = False
+        removed_nets = set()
+        for net in invert_destination_wires.values():
+            # If the argument of the net is in invert_destination_wires, then it is a double invert
+            # If the net is in net_exclude_set, then it was already removed, so we do not process it
+            if net.args[0].name in invert_destination_wires and net not in net_exclude_set:
+                previous_net = invert_destination_wires[net.args[0].name]
```
fdxmw: This repeats the dictionary lookup from the previous line, we can avoid one lookup with something like

```
previous_net = invert_destination_wires.get(net.args[0].name)
if previous_net is not None and ...
```

I think this also lets you combine this `if` statement with the next one

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+                    return True
+        return False
+
+    new_logic = set()
+    net_exclude_set = set()  # removed nets
+    wire_removal_set = set()
+    # Dictionary, key is the destination wire of the invert net, value is the invert net
+    invert_destination_wires = {}
+    for net in block.logic:
+        if net.op == "~":
```
fdxmw: Can we remove the `used_elsewhere` check if we check here for multiple destinations? Something like
```
if net.op == "~" and len(net.dests) == 1:
```

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    def is_wirevector_used_elsewhere(wire, used_at_nets):
+        for net in block.logic:
+            if net not in used_at_nets:
+                if wire.name in [x.name for x in net.args] \
+                        or wire.name in [x.name for x in net.dests]:
+                    return True
+        return False
+
+    new_logic = set()
+    net_exclude_set = set()  # removed nets
```
fdxmw: Seems better to call this `removed_nets` and remove the comment? :)

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+                    # was not removed yet. If it was not yet removed, the for loop
+                    # needs to run again, so we set repeat to True.
+                    if previous_net.args[0].name in invert_destination_wires:
+                        repeat = True
+                    else:
+                        new_logic.add(LogicNet('w', None, args=previous_net.args, dests=net.dests))
+                        wire_removal_set.add(net.args[0])
+                        removed_nets.add(net)
+                        removed_nets.add(previous_net)
+        # remove removed_nets from invert_destination_wires to optimize the for loop
```
fdxmw: I think this is needed to ensure the loop eventually stops (it's not just an optimization?) since line 106 will keep `repeat`ing until the net is removed from `invert_destination_wires`

pyrtl/passes.py
```diff
@@ -52,11 +52,84 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+                    # If previous_net is in invert_destination_wires, we have a chain of
+                    # 3 or more double inverts. To make sure we remove double inverts
+                    # in these chains sequentially, we only remove the double invert
+                    # we found if the invert net whose destination is previous_net
+                    # was not removed yet. If it was not yet removed, the for loop
+                    # needs to run again, so we set repeat to True.
+                    if previous_net.args[0].name in invert_destination_wires:
+                        repeat = True
+                    else:
+                        new_logic.add(LogicNet('w', None, args=previous_net.args, dests=net.dests))
```
fdxmw: This adds a wire net, which can't be removed by `_remove_wire_nets` since that pass ran already in `optimize()`. 

It seems be possible to do this without adding another wire, see how `_remove_wire_nets` rewrites `net.args` for other nets. I think we could do something similar here, with a data structure like `_ProducerList` that tracks even-numbered chains of inverters, and identifies the ultimate source of these chains

fdxmw (commented):
Please start with the comments on line 109 and 81 first. If those can be addressed, they will probably make many other comments obsolete. Thank you for improving this code!

gaborszita:
I will fix these next week, I was busy this weekend.

fdxmw:
> I will fix these next week, I was busy this weekend.

Sounds great, no rush! Let me know if you have any questions about my comments, happy to discuss.

pyrtl/passes.py
gaborszita: Yes, `_remove_wire_nets` was already called in optimize(), but that's why I call `_remove_wire_nets` again at the end of this function to remove the extra added wire nets. When we export the net list to an svg, it shows the extra wire nets and you can see in the svgs I provided that there aren't any. But I'll see if there is a more efficient way to do this to avoid calling `_remove_wire_nets` twice.

pyrtl/passes.py
gaborszita: I tried that first, but that requires comparing wirenets. Since the invert_destination_wires would hold wirevector objects, not their names, the `is_wirevector_used_elsewhere(wire, used_at_nets)` internal function would become this:

```
def is_wirevector_used_elsewhere(wire, used_at_nets):
        for net in block.logic:
            if net not in used_at_nets:
                if wire in [x for x in net.args] \
                        or wire in [x for x in net.dests]:
                    return True
        return False
```

The `in` operator in Python uses the `=` operator to check equivalency. However, this throws an exception, because the `=` operator is not supported on WireVectors (the `=` operator gets executed using the `__eq__` special method, which calls the `_two_var_op` method, which throws the exception):

```
Traceback (most recent call last):
  File "/Users/gabi/PyRTL/mytest.py", line 18, in <module>
    pyrtl.optimize()
    ~~~~~~~~~~~~~~^^
  File "/Users/gabi/PyRTL/pyrtl/passes.py", line 55, in optimize
    _remove_double_inverts(block, skip_sanity_check)
    ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gabi/PyRTL/pyrtl/passes.py", line 87, in _remove_double_inverts
    if not is_wirevector_used_elsewhere(net.args[0], (net, previous_net)) \
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gabi/PyRTL/pyrtl/passes.py", line 68, in is_wirevector_used_elsewhere
    if wire in [x for x in net.args] \
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/gabi/PyRTL/pyrtl/wire.py", line 401, in __eq__
    return self._two_var_op(other, '=')
           ~~~~~~~~~~~~~~~~^^^^^^^^^^^^
  File "/Users/gabi/PyRTL/pyrtl/wire.py", line 241, in _two_var_op
    working_block().add_net(net)
    ~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/gabi/PyRTL/pyrtl/core.py", line 375, in add_net
    self.sanity_check_net(net)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/Users/gabi/PyRTL/pyrtl/core.py", line 800, in sanity_check_net
    raise PyrtlInternalError('error, net op "%s" not from acceptable set %s' %
                             (net.op, self.legal_ops))
pyrtl.pyrtlexceptions.PyrtlInternalError: error, net op "=" not from acceptable set {'&', '@', '|', '^', '~', 'r', 's', 'w', 'm', 'c', 'n'}
```

So, I decided to use the names of the WireVectors as those are guaranteed to be unique.

pyrtl/passes.py
gaborszita: Oh yes, I didn’t realize I could iterate over the `block` instead of `block.logic` to get topological order instead of the random order of elements in `block.logic`. I’ve addressed this and pushed the changes. My program now passes my tests for the double inverts, but it fails an unrelated test, `test_loop_1 (tests.test_helperfuncs.TestLoopDetection.test_loop_1)`, with the error:

`pyrtl.pyrtlexceptions.PyrtlError: Failure in Block Iterator due to non-register loops`

I’m not very familiar with the internal workings of PyRTL, so I don’t fully understand why I’m getting this error. Is iterating over non-register loops disallowed? If so, what’s the best way to fix this? I’ve also noticed that other methods in `optimize.py` iterate over `block.logic` instead of `block` — is this specifically to avoid this error?

pyrtl/passes.py
gaborszita: I named it `net_exclude_set` instead of something like `net_removed_set` because the nets added to it aren’t removed immediately from the logic net - they are excluded later when creating the new logic net `new_logic`. But I agree, the naming is a little confusing. I’ll probably update the comment to something like `Nets to remove, these will not be included when creating the new logic net.` I will fix these smaller issues after we address the bigger issues.

pyrtl/passes.py
gaborszita: I am aware of the tests GitHub check failing, I pushed this commit intentionally so it's easy for you to see which test that is failing.

pyrtl/passes.py
gaborszita: I tried a simpler method to check if an invert net is used elsewhere but wasn't able to come up with any good solutions. The issue with what you're suggesting, as far as I understand PyRTL, is that other nets don't connect directly to an invert net - they connect to a WireVector, which then connects to the invert net. For example, if `N` is an invert net and net `A` connects to `N`, `A` doesn't actually have its destination set to `N`, and `N` doesn't have its source set to `A` (except if A is an `Input` or `Output` net). Instead, A's destination is a WireVector, and N's source is that WireVector. However, WireVectors don’t store their sources and destinations. The only method I’ve found to found the sources and destinations of a WireVector was to manually search for them through all the nets.

pyrtl/passes.py
gaborszita: The idea is that when the program reaches the top of an inverter chain (which means all inverters in the chain have been processed), `previous_net.args[0]` won't be an inverter net. Hence, `previous_net.args[0].name in invert_destination_wires` will be false. I just checked and removing this optimization doesn't cause an infinite while loop.

pyrtl/passes.py
fdxmw: You are correct, sorry I forgot that an inverter LogicNet will always only have one `dest`. But I still think you can simplify checking for other users with [block.net_connections](https://pyrtl.readthedocs.io/en/latest/blocks.html#pyrtl.core.Block.net_connections), which lets you efficiently map from a WireVector to the LogicNet that produces it, and from a WireVector to the LogicNets that use it. Something like this:

```
>>> i = pyrtl.Input(bitwidth=2, name="i")
>>> o1 = pyrtl.Output(bitwidth=2, name="o1")
>>> o2 = pyrtl.Output(bitwidth=2, name="o2")
>>> multiuser = pyrtl.WireVector(bitwidth=2, name="multiuser")
>>> multiuser <<= i
>>> o1 <<= multiuser
>>> o2 <<= multiuser

>>> print(pyrtl.working_block())
multiuser/2W <-- w -- i/2I
o1/2O <-- w -- multiuser/2W
o2/2O <-- w -- multiuser/2W

>>> _, users = pyrtl.working_block().net_connections()
>>> len(users[multiuser])
2
>>> len(users[i])
1
```

pyrtl/passes.py
fdxmw: You can put WireVectors in sets and dicts and check for their presence with `in`:

```
>>> w1 = pyrtl.WireVector(name="w1")
>>> w2 = pyrtl.WireVector(name="w2")

>>> s = {w1}
>>> w1 in s
True
>>> w2 in s
False

>>> d = {w1: 1}
>>> w1 in d
True
>>> w2 in d
False
```
I believe it works because `in` is equivalent to `any(x is e or x == e for e in y)` (see [reference](https://docs.python.org/3/reference/expressions.html#membership-test-operations), which checks for `is` before checking `==`, and sets and dicts use `__hash__` to skip equality checks.

You're right that you can't check if a WireVector is in a list because of the issue you point out with the `__eq__` overload, which is confusing, but it's generally better to check if a WireVector is in a set rather than a list because the lookup is O(1) instead of O(n).

pyrtl/passes.py
fdxmw: Hmm, it looks like topological-order iteration throws an exception if there are combinational loops in the block ([reference](https://github.com/UCSBarchlab/PyRTL/blob/5b282f367baffa2e434960652419f2b4d856d405/pyrtl/core.py#L599)), which is reasonable, but unfortunate because we have some `TestLoopDetection` tests that create combinational loops then call `optimize` (see [assert_has_loop](https://github.com/UCSBarchlab/PyRTL/blob/development/tests/test_helperfuncs.py#L1178), which seems to trigger the test failures you're seeing.

I think there are a few options:
1. It should be possible to write this optimization as a constant number of passes over the block without relying on topological iteration order, but the code will be a bit more complicated.
2. Change the iterator's behavior so it ignores combinational loops, and doesn't throw an exception.
3. Change `TestLoopDetection` so it doesn't try to optimize combinational loops.
4. Catch the exception, and skip the inverter optimization if the block contains a combinational loop. This is the quick and dirty solution, but these solutions are usually fragile since it's always hard to tell exactly who threw the exception. You can get more information by inspecting the exception itself, but that inevitably leads to string matching and regular expressions, and now you have two problems :)

I'd recommend option 1 for now. I'll check with Tim about options 2 and 3. I don't recommend option 4.

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+        for inverter_chain in inverter_chains:
+            chain_arg = inverter_chain[0]
+            chain_dest = inverter_chain[-1]
+            chain_users = len(wire_users[chain_dest])
+
+            if chain_dest is current_arg and chain_users == 1:
+                # This chain's only destination is the current inverter. Append the
+                # current inverter to the chain.
+                append_to = inverter_chain
+            elif chain_arg is current_dest and current_users <= 1:
```
fdxmw: I think this can check for `current_users == 1` rather than `current_users <= 1` because `chain_arg is current_dest` implies there is at least one user. And we only want to prepend the current inverter if the current inverter has exactly one user, so `current_users == 1` makes that intent more clear

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+                # This chain's only argument is the current inverter. Add the current
+                # inverter to the beginning of the chain.
+                prepend_to = inverter_chain
+            else:
+                next_inverter_chains.append(inverter_chain)
```
fdxmw, line 110: Consider adding a comment on what this does - if we reach this block, the current inverter does not interact with `inverter_chain`, so we pass the `inverter_chain` through to `next_inverter_chains`

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            # The current inverter joins two existing inverter chains.
+            next_inverter_chains.append(append_to + prepend_to)
+        elif append_to:
+            # Add the current inverter after 'append_to'.
+            next_inverter_chains.append(append_to + [current_dest])
+        elif prepend_to:
+            # Add the current inverter before 'prepend_to'.
+            next_inverter_chains.append([current_arg] + prepend_to)
+        else:
+            next_inverter_chains.append([current_arg, current_dest])
```
fdxmw, line 124: Consider adding a comment on what this does - the current inverter did not interact with any existing `inverter_chain` so it starts a new inverter chain

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+
+        inverter_chains = next_inverter_chains
+
+    new_logic = set()
```
fdxmw, line 151: This function is quite long, so consider moving the code that builds `inverter_chains` into a helper function, since the rest of the code only seems to use `inverter_chains`

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    net_removal_set = set()
+    wire_removal_set = set()
+    for inverter_chain in inverter_chains:
+        if len(inverter_chain) > 2:
```
fdxmw, line 179: Add some more comments about what `len(inverter_chain)` means - it's a little counterintuitive that a length-2 `inverter_chain` represents a single inverter, and a length-3 `inverter_chain` represents two chained inverters

pyrtl/passes.py
```diff
@@ -52,11 +52,100 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            if len(inverter_chain) % 2 == 1:  # even number of inverters in a chain
+                end_idx = len(inverter_chain) - 1
+            else:  # odd number of inverters in a chain
+                end_idx = len(inverter_chain) - 2
+            wires_to_remove = inverter_chain[1:end_idx]
+            new_logic.add(LogicNet('w', None, args=(inverter_chain[0],),
```
fdxmw: Try writing this pass so we don't create new wires, then immediately remove them on line 146 below. It seems possible to rewrite the block, updating any LogicNets that use the end of an inverter chain to instead use the beginning of that inverter chain. The code should be similar to the second pass in `_remove_wire_nets` [here](https://github.com/UCSBarchlab/PyRTL/blob/development/pyrtl/passes.py#L92)

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    # wire_creator maps from WireVector to the LogicNet that defines its value.
+    # wire_users maps from WireVector to a list of LogicNets that use its value.
+    wire_creator, wire_users = block.net_connections()
+
+    new_logic = set()
+    net_removal_set = set()
+    wire_removal_set = set()
+    wire_src_dict = _ProducerList()
+    for inverter_chain in _get_inverter_chains(wire_creator, wire_users):
+        # if len(inverter_chain) = n, there are n-1 inverters in the chain
```
fdxmw: > Comments should be complete sentences. The first word should be capitalized, unless it is an identifier that begins with a lower case letter (never alter the case of identifiers!).

https://peps.python.org/pep-0008/#comments

Please try to use complete sentences in comments throughout this pull request. Note that "complete sentences" also implies that comments should end with punctuation :)

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    """ Removes all double invert nets from the block. """
+
+    # wire_creator maps from WireVector to the LogicNet that defines its value.
+    # wire_users maps from WireVector to a list of LogicNets that use its value.
+    wire_creator, wire_users = block.net_connections()
+
+    new_logic = set()
+    net_removal_set = set()
+    wire_removal_set = set()
+    wire_src_dict = _ProducerList()
```
fdxmw, line 174: `_ProducerList` seems like overkill here since there should be exactly one hop from the end of an inverter chain to the beginning of the inverter chain. If that's right, it might be clearer to use a `dict`, with `dict.get` to set the appropriate default value when a WireVector is not in the dict https://docs.python.org/3/library/stdtypes.html#dict.get 

Someone reading this code is much more likely to know how `dict` works, and much less likely to know how `_ProducerList` works

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+    for inverter_chain in _get_inverter_chains(wire_creator, wire_users):
+        # if len(inverter_chain) = n, there are n-1 inverters in the chain
+        # only remove inverters if there are at least two inverters in a chain
+        if len(inverter_chain) > 2:
+            if len(inverter_chain) % 2 == 1:  # even number of inverters in a chain
+                end_idx = len(inverter_chain) - 1
```
fdxmw: I think this code might be a little cleaner if we keep the first inverter in odd-length chains rather than the last inverter? So we'd track `start_idx` (which would always be just `1` or `2`) instead of `end_idx` (no more `len(inverter_chain) - 1`, etc), and I think you can do `wires_to_remove = inverter_chain[start_idx:]` (no more +1 adjustment for `end_idx`)

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            wires_to_remove = inverter_chain[1:end_idx + 1]
+            wire_removal_set.update(wires_to_remove)
+            # remove inverters used in the chain
+            inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
+            inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
+            net_removal_set.update(inverters_to_remove)
+            # map the end wire of the inverter chain to the beginning wire
+            wire_src_dict[inverter_chain[end_idx]] = inverter_chain[0]
+
+    for net in block.logic:
```
fdxmw, line 197: This loop is doing a lot of work so add a comment on what it's doing - I think this is rewriting the block's logic, skipping unnecessary inverters, and directly connecting any removed inverter chain's input to its users

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            if len(inverter_chain) % 2 == 1:  # even number of inverters in a chain
+                end_idx = len(inverter_chain) - 1
+            else:  # odd number of inverters in a chain
+                end_idx = len(inverter_chain) - 2
+            # remove wires used in the inverter chain
+            wires_to_remove = inverter_chain[1:end_idx + 1]
+            wire_removal_set.update(wires_to_remove)
+            # remove inverters used in the chain
+            inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
+            inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
```
fdxmw: Is this necessary? It looks like `inverter_chain[end_idx]` is in `wires_to_remove`, so it should already be in the set

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
         _remove_unlistened_nets(block)
         common_subexp_elimination(block)
+        _remove_double_inverts(block, skip_sanity_check)
         if (not skip_sanity_check) or _get_debug_mode():
             block.sanity_check()
     return block
 
 
+def _get_inverter_chains(wire_creator, wire_users):
+    """Returns all inverter chains in the block"""
```
fdxmw: We should describe this function's return value (structure and semantics) in much more detail here. If you find yourself writing redundant comments, I'd move the comments to here.

pyrtl/passes.py
```diff
@@ -52,11 +52,113 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            # The current inverter is not connected to any inverter chain, so
+            # we start a new inverter chain with it
+            next_inverter_chains.append([current_arg, current_dest])
+
+        inverter_chains = next_inverter_chains
+    return inverter_chains
+
+
+def _remove_double_inverts(block, skip_sanity_check=False):
+    """ Removes all double invert nets from the block. """
```
fdxmw: Add more details to this comment - what is a double invert, what about triple/quadruple/etc inverts, what if an intermediate value has another user, ...

fdxmw (commented):
Overall structure of this pull request looks great, so we are on to fine details at this point. We're almost there, thank you for your patience!

pyrtl/passes.py
gaborszita: That's what I thought too, so initially I used a `dict`, please see commit `99c3510`. However, the `test_multiple_double_invert_chains` test, along with multiple other tests, failed with this implementation.
Consider following logic:

- There are four main WireVectors A, B, and C
- A = ~(~B)
- B = ~(~C)

During optimization, the `A-B` and `B-C` double inverter chains will be optimized, and the `B` and `C` WireVectors will be marked for removal. `B` will be linked `A`  and `C` will be linked to `B` in the wire replacement dict. The problem arises when recreating the block logic: When using just a `dict`, the new source for a `LogicNet` using `C` as its source will be `B` - but this is obviously wrong, since `B` was removed! The source for a LogicNet using `C` as its source should be `A`.

If a `WireVector` has a replacement in the wire replacement dict, we need to check the dict again because this replacement `WireVector` may also have been replaced by another `WireVector`, and continue until we find a `WireVector` that has not been replaced - and this is exactly what `_ProducerList` does.

fdxmw: > this replacement WireVector may also have been replaced by another WireVector

Hmm this doesn't quite make sense to me - if the replacement WireVector was itself replaced by another WireVector, why are there two separate inverter chains? We should have built one long inverter chain instead of two short ones?

gaborszita: Not if one intermediate `WireVector` has another user (should've been more clear about this, sorry). For example, here, even though there are 4 inverters in a chain, 2 separate inverter chains will be recognized because the `WireVector` after the second inverter has another user.



The wire replacement dict for this example would be:
- 0x4e0 -> 0x9d0
- 0x9d0 -> 0x110

The added arrows in this picture show how the wires are linked in the replacement dict:



So the `WireVectors` 0x4e0 and 0x9d0 would both be replaced by the `WireVector` 0x110 (but we need to iterate twice over the the dict when finding the replacement for 0x4e0), producing this optimized output:

fdxmw: Ah, thank you for the detailed explanation! I agree with your analysis, but please add some comments explaining why we need a `_ProducerList` here. It would help to add this example in the comments, maybe we could add a simple diagram, something like
```
A -~-> B -~-> C -w-> X
C -~-> D -~-> E -w-> Y
```
which tries to show the two double-inverter chains (`AC`, `CE`), and how `C` has two users. This should optimize to
```
A -w-> X
A -w-> Y
```
(edited: my arrows were backwards :(

pyrtl/passes.py
gaborszita: The function's return value? It is a void function, it doesn't return anything.

gaborszita:
Sorry for not squashing my commits, I forgot I already had commits locally when I pushed.

pyrtl/passes.py
fdxmw: I think this returns `inverter_chains` on line 116 below? Regardless, the intent of my comment was that this comment should define what an inverter chain is, and how it's represented in the code.

gaborszita: Oh yes, sorry, I was looking at a different function, I'll fix this.

pyrtl/passes.py
```diff
@@ -116,8 +116,22 @@ def _get_inverter_chains(wire_creator, wire_users):
...
-    """ Removes all double invert nets from the block. """
+def _remove_inverter_chains(block, skip_sanity_check=False):
+    """ Removes all inverter chains from the block.
+
+    An inverter chain means two or more inverters directly connected
+    to each other. Inverter chains are redundant and can be removed.
+    For example, A -~-> B -~-> C -w-> X can be reduced to A -w-> X.
+
+    After optimization, a chain of an even number of inverters will
+    be reduced a direct connection, and a chain of an odd number of
```
fdxmw, line 138: A chain with an odd number of inverters will be reduced to one inverter, which does not really remove the inverter chain. So maybe this function should be called `_optimize_inverter_chains`?

pyrtl/passes.py
```diff
@@ -126,25 +140,49 @@ def _remove_double_inverts(block, skip_sanity_check=False):
...
+    # We need a ProducerList, rather than a simple dict, because if an inverter chain
+    # of more than two inverters has intermediate users, we may have to query the dict
+    # multiple times to get the replacement for the inverter chain's last wire.
+    # Consider the following circuit, for example:
+    # A -~-> B -~-> C -w-> X
+    # C -~-> D -~-> E -w-> Y
+    # This is the optimized version of the circuit:
+    # A -w-> X
+    # A -w-> Y
+    # The inverter chains found will be B-C and D-E (two separate chains will be
```
fdxmw: It might be better to call these chains `A-B-C` and `C-D-E` because that might make it more clear that the chains include all the inverters in the example (with `B-C`, it's less clear if the chain includes the inverter between `A` and `B`)

pyrtl/passes.py
```diff
@@ -126,25 +140,49 @@ def _remove_double_inverts(block, skip_sanity_check=False):
...
+    # found instead of B-C-D-E because C has an intermediate user). In the dict,
+    # C will be mapped to A and E will be maped to C. Hence, when finding the
```
fdxmw: Typo: mapped

pyrtl/passes.py
```diff
@@ -126,25 +140,49 @@ def _remove_double_inverts(block, skip_sanity_check=False):
...
+            # Remove inverters used in the chain.
             inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
-            inverters_to_remove.add(wire_creator[inverter_chain[end_idx]])
             net_removal_set.update(inverters_to_remove)
-            # map the end wire of the inverter chain to the beginning wire
-            wire_src_dict[inverter_chain[end_idx]] = inverter_chain[0]
+            # Map the end wire of the inverter chain to the beginning wire.
+            wire_src_dict[inverter_chain[-1]] = inverter_chain[start_idx - 1]
 
+    # This loop recreates the LogicNet with inverter chains removed. It adds each
```
fdxmw: I think the words `LogicNet` and `block` are swapped in this comment - we're recreating the block, and adding each LogicNet in the original block to the new block ...

pyrtl/passes.py
```diff
@@ -52,11 +52,162 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
         if (not skip_sanity_check) or _get_debug_mode():
             block.sanity_check()
     return block
 
 
+def _get_inverter_chains(wire_creator, wire_users):
+    """Returns all inverter chains in the block.
+
+    The function returns a list of inverter chains in the block.
+    Each inverter chain is represented as a list of the LogicNets
```
fdxmw: I believe it's a list of the WireVectors in the chain? (not LogicNets)

pyrtl/passes.py
```diff
@@ -52,11 +52,162 @@ def optimize(update_working_block=True, block=None, skip_sanity_check=False):
...
+            wire_removal_set.update(wires_to_remove)
+            # Remove inverters used in the chain.
+            inverters_to_remove = {wire_creator[wire] for wire in wires_to_remove}
+            net_removal_set.update(inverters_to_remove)
+            # Map the end wire of the inverter chain to the beginning wire.
+            wire_src_dict[inverter_chain[-1]] = inverter_chain[start_idx - 1]
+
+    # This loop recreates the block with inverter chains removed. It adds each
+    # LogicNet in the original block to the new block if it is not marked for
+    # removal, and replaces the source of the LocigNet if its source was the end wire
```
fdxmw: Typo: `LogicNet`

fdxmw (commented):
This should be the last round of comments, thank you for your patience!

fdxmw (approved):
Looks great, thank you for all your effort on this pull request!

PR #464, opened 2025-02-19
Allow negative Register reset values

PrajwalVandana:
Pass bitwidth to `infer_val_and_bitwidth` in Register.__init__ to allow negative Register reset values

### DIFF
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -804,8 +804,8 @@ def _convert_verilog_str(val: str, bitwidth: int = None,
         if verilog_bitwidth > bitwidth:
             raise PyrtlError(
                 "bitwidth parameter passed (%d) cannot fit Verilog-style constant with bitwidth %d"
-                % (bitwidth, verilog_bitwidth) +
-                " (if bitwidth=None is used, PyRTL will determine the bitwidth from the "
+                % (bitwidth, verilog_bitwidth)
+                + " (if bitwidth=None is used, PyRTL will determine the bitwidth from the "
                 "Verilog-style constant specification)"
             )
 
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -789,7 +789,6 @@ def _convert_verilog_str(val: str, bitwidth: int = None,
         raise PyrtlError('error, "signed" option with verilog-style string constants not supported')
 
     bases = {'b': 2, 'o': 8, 'd': 10, 'h': 16, 'x': 16}
-    passed_bitwidth = bitwidth
 
     neg = False
     if val.startswith('-'):
@@ -800,18 +799,19 @@ def _convert_verilog_str(val: str, bitwidth: int = None,
     if len(split_string) != 2:
         raise PyrtlError('error, string not in verilog style format')
     try:
-        bitwidth = int(split_string[0])
-        if passed_bitwidth is not None:
-            if bitwidth > passed_bitwidth:
-                raise PyrtlError(
-                    "bitwidth parameter passed (%d) cannot fit Verilog-style constant with bitwidth %d"
-                        % (passed_bitwidth, bitwidth)
-                )
-            bitwidth = passed_bitwidth
+        verilog_bitwidth = int(split_string[0])
+        bitwidth = bitwidth or verilog_bitwidth  # if bitwidth is None, use verilog_bitwidth
+        if verilog_bitwidth > bitwidth:
+            raise PyrtlError(
+                "bitwidth parameter passed (%d) cannot fit Verilog-style constant with bitwidth %d"
+                % (bitwidth, verilog_bitwidth) +
+                " (if bitwidth=None is used, PyRTL will determine the bitwidth from the "
+                "Verilog-style constant specification)"
+            )
 
         sval = split_string[1]
         if sval[0] == 's':
-            raise PyrtlError('error, signed integers are not supported in verilog-style constants')
+            raise PyrtlError('error, signed integers are not supported in Verilog-style constants')
         base = 10
         if sval[0] in bases:
             base = bases[sval[0]]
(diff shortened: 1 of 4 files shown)


fdxmw:
Thank you for your contribution! This change looks good to me, but you'll need to fix the test failures before we can merge it. Please also add a test for the new behavior so we can make sure it doesn't regress. Thanks again!

pyrtl/wire.py
```diff
@@ -852,7 +852,11 @@ def __init__(self, bitwidth: int, name: str = '', reset_value: int = None,
         super(Register, self).__init__(bitwidth=bitwidth, name=name, block=block)
         self.reg_in = None  # wire vector setting self.next
         if reset_value is not None:
-            reset_value, rst_bitwidth = infer_val_and_bitwidth(reset_value)
+            # NOTE: passing bitwidth here to allow for negative reset values
+            reset_value, rst_bitwidth = infer_val_and_bitwidth(
+                reset_value,
+                bitwidth=(None if isinstance(reset_value, str) else bitwidth)
```
fdxmw: Is it possible to avoid this special case for `str`? Ideally that should be handled for us by `infer_val_and_bitwidth`. What you had before seems better to me (`bitwidth=bitwidth`).

I think this special case can introduce confusing behavior when a Verilog-style `reset_value` is negative with a bitwidth smaller than the `Register`'s bitwidth - try `pyrtl.Register(bitwidth=4, reset_value="-3'd1")` The `reset_value` becomes 7 (not -1) because the smaller reset value is zero-extended (not sign-extended) up to the register's bitwidth.

If you added this just to fix `TestRegister.test_reset_value_as_string`, I think it's better to fix that test so its `Register`'s `reset_value` matches its `bitwidth`, but let me know if you have other ideas.

This pull request may require users to change their code, since anyone that uses a smaller Verilog-style `reset_value` will need to update their code so the bitwidths match. To support signed `reset_value`s we can't safely zero-extend smaller `reset_value`s anymore. If that sounds right to you, please add a note about this to `CHANGELOG.md`, in a `## [Unreleased]` section.

tests/test_wire.py
```diff
@@ -263,13 +263,30 @@ def test_reset_value_as_string(self):
         self.assertEqual(r.reset_value, 1)
 
     def test_invalid_reset_value_too_large(self):
-        with self.assertRaisesRegex(pyrtl.PyrtlError, "cannot fit in the specified"):
+        with self.assertRaisesRegex(
+            pyrtl.PyrtlError,
+            "bitwidth specified \\(4\\) is insufficient to represent constant 16"
```
fdxmw: As you've seen, error message checks in tests can be pretty annoying because they can fail to pattern-match after a correct code change. Consider removing this error message check (just check that a `PyrtlError` was raised), or generalizing the error message pattern so it's at least not sensitive to the exact values selected below (4, 16)

tests/test_wire.py
```diff
@@ -263,13 +263,30 @@ def test_reset_value_as_string(self):
...
 
+    def test_negative_reset_value(self):
+        r = pyrtl.Register(4, reset_value=-4)
+        self.assertEqual(
+            pyrtl.helperfuncs.val_to_signed_integer(r.reset_value, r.bitwidth),
+            -4
+        )
+
+    def test_negative_reset_value_as_string(self):
+        r = pyrtl.Register(4, reset_value="-4'd1")
```
fdxmw, line 281: Also add tests for when a Verilog-style `reset_value`'s bitwidth is larger or smaller than the `Register`'s bitwidth

pyrtl/wire.py
PrajwalVandana: If the Verilog string is a negative literal, can we not "safely" sign-extend it? That seems like it would fix this problem as well

fdxmw: That sounds good to me, assuming we can keep all the extension logic contained within `_convert_verilog_str`

fdxmw: Please try to merge this logic into `_convert_verilog_str` ([reference](https://github.com/UCSBarchlab/PyRTL/blob/development/pyrtl/helperfuncs.py#L786)), which is called by `infer_val_and_bitwidth`. We should not detect and parse Verilog string constants here, that should all be handled by `infer_val_and_bitwidth` and its helpers

pyrtl/helperfuncs.py
```diff
@@ -795,11 +795,20 @@ def _convert_verilog_str(val: str, bitwidth: int = None,
...
         raise PyrtlError('error, string not in verilog style format')
     try:
         bitwidth = int(split_string[0])
+        if passed_bitwidth is not None:
+            if bitwidth > passed_bitwidth:
+                raise PyrtlError(
+                    "bitwidth parameter passed (%d) cannot fit Verilog-style constant with bitwidth %d"
+                        % (passed_bitwidth, bitwidth)
+                )
+            bitwidth = passed_bitwidth
```
fdxmw: This is the inverse of line 792 above :) I think we can delete lines 792 and 810 if we change line 803 to `verilog_bitwidth = int(split_string[0])`, since the Verilog bitwidth doesn't seem to be used outside of this code block. 

I think this also lets you combine the two `if` statements on lines 804 and 805 into `if bitwidth is not None and verilog_bitwidth > bitwidth:`

pyrtl/helperfuncs.py
```diff
@@ -795,11 +795,20 @@ def _convert_verilog_str(val: str, bitwidth: int = None,
...
+
     split_string = val.lower().split("'")
     if len(split_string) != 2:
         raise PyrtlError('error, string not in verilog style format')
     try:
         bitwidth = int(split_string[0])
+        if passed_bitwidth is not None:
+            if bitwidth > passed_bitwidth:
+                raise PyrtlError(
+                    "bitwidth parameter passed (%d) cannot fit Verilog-style constant with bitwidth %d"
```
fdxmw: I think we should keep the `(if bitwidth=None is used, PyRTL will determine the bitwidth from the Verilog-style constant specification)` hint in this error message

fdxmw (commented):
Thank you for your patience, we are almost there, these should be my last comments :)

There are also a couple pycodestyle errors on this pull request that will need to be fixed:
  pyrtl/helperfuncs.py:807:101: E501 line too long (103 > 100 characters)
  pyrtl/helperfuncs.py:808:25: E131 continuation line unaligned for hanging indent

PR #466, opened 2025-02-21
Add helper function for binary to 1-hot

### LINKED ISSUE #465: Add a helper function to encode binary to 1-hot
Takes an input representing a bit position and returns a WireVector with that bit position set to 1 and the others to 0.

vaaniarora:
Takes an input representing a bit position and returns a WireVector with that bit position set to 1 and the others to 0.

Closes #465

### DIFF
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -13,7 +13,15 @@ from typing import Union, NamedTuple
 from .core import working_block, _NameIndexer, _get_debug_mode, Block
 from .pyrtlexceptions import PyrtlError, PyrtlInternalError
 from .wire import WireVector, Input, Output, Const, Register, WrappedWireVector
-from .corecircuits import as_wires, rtl_all, rtl_any, concat, concat_list, select
+from .corecircuits import (
+    as_wires,
+    rtl_all,
+    rtl_any,
+    concat,
+    concat_list,
+    select,
+    shift_left_logical
+)
 
 # -----------------------------------------------------------------
 #        ___       __   ___  __   __
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
         already_found = already_found | w[i]
 
     return pos
+
+
+def binary_to_one_hot(bit_position, max_bitwidth: int = None) -> WireVector:
+    '''Takes an input representing a bit position and returns a WireVector
+    with that bit position set to 1 and the others to 0.
+
+    :param bit_position: WireVector, WireVector-like object, or something that can be converted
+        into a :py:class:`.Const` (in accordance with the :py:func:`.as_wires()`
+        required input). Example inputs: ``0b10``, ``0b1000, ``4``.
+    :param max_bitwidth: Optional integer maximum bitwidth for the resulting one-hot WireVector.
+    :return: WireVector with the bit position, given by the input, set to 1 and all others set to 0.
+
+    If the max_bitwidth provided is not sufficient for the given bit_position to be set to 1,
+    a ``0`` WireVector of size max_bitwidth will be returned.
+
+    Examples::
+
+        binary_to_onehot(2)  # returns 0b0100
+        binary_to_onehot(8)  # returns 0b00100000
+        binary_to_onehot(12)  # returns 0b0001000000000000
+        binary_to_onehot(15)  # returns 0b1000000000000000
+
+    '''
+
+    bit_position = as_wires(bit_position)
+
+    if max_bitwidth is not None:
+        bitwidth = max_bitwidth
+    else:
+        bitwidth = 2 ** len(bit_position)
+
+    onehot = Const(1, bitwidth=bitwidth)
+
+    return shift_left_logical(onehot, bit_position)
--- a/pyrtl/helperfuncs.py
+++ b/pyrtl/helperfuncs.py
@@ -1731,20 +1731,19 @@ def binary_to_one_hot(bit_position, max_bitwidth: int = None) -> WireVector:
 
     :param bit_position: WireVector, WireVector-like object, or something that can be converted
         into a :py:class:`.Const` (in accordance with the :py:func:`.as_wires()`
-        required input). Example inputs: ``0b10``, ``0b1000, ``4``.
+        required input). Example inputs: ``0b10``, ``0b1000``, ``4``.
     :param max_bitwidth: Optional integer maximum bitwidth for the resulting one-hot WireVector.
-    :return: WireVector with the bit position, given by the input, set to 1 and all others set to 0.
+    :return: WireVector with the bit position given by the input set to 1 and all other bits
+    set to 0 (bit position 0 being the least significant bit).
 
     If the max_bitwidth provided is not sufficient for the given bit_position to be set to 1,
     a ``0`` WireVector of size max_bitwidth will be returned.
 
     Examples::
 
-        binary_to_onehot(2)  # returns 0b0100
-        binary_to_onehot(8)  # returns 0b00100000
-        binary_to_onehot(12)  # returns 0b0001000000000000
-        binary_to_onehot(15)  # returns 0b1000000000000000
-
+        binary_to_onehot(0)  # returns 0b01
+        binary_to_onehot(3)  # returns 0b1000
+        binary_to_onehot(0b100)  # returns 0b10000
     '''
 
     bit_position = as_wires(bit_position)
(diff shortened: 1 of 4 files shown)


pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
...
     return pos
+
+
+def binary_to_one_hot(bit_position, max_bitwidth: int = None) -> WireVector:
+    '''Takes an input representing a bit position and returns a WireVector
+    with that bit position set to 1 and the others to 0.
+
+    :param bit_position: WireVector, WireVector-like object, or something that can be converted
+        into a :py:class:`.Const` (in accordance with the :py:func:`.as_wires()`
+        required input). Example inputs: ``0b10``, ``0b1000, ``4``.
```
fdxmw: We're missing two backquotes after `0b1000` here, see the generated docs for this pull request at https://pyrtl--466.org.readthedocs.build/en/466/helpers.html#pyrtl.helperfuncs.binary_to_one_hot . You can find this link by clicking on the "5 successful checks" dropdown, then clicking on the readthedocs link

pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
...
+    :param max_bitwidth: Optional integer maximum bitwidth for the resulting one-hot WireVector.
+    :return: WireVector with the bit position, given by the input, set to 1 and all others set to 0.
+
+    If the max_bitwidth provided is not sufficient for the given bit_position to be set to 1,
+    a ``0`` WireVector of size max_bitwidth will be returned.
+
+    Examples::
+
+        binary_to_onehot(2)  # returns 0b0100
+        binary_to_onehot(8)  # returns 0b00100000
```
fdxmw: This example seems wrong? The one seems to be in bit position 5, not 8. Also I think it would be a little better to use smaller examples so it's easier to see how this works, without having to carefully count zeroes :)

pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
...
+    bit_position = as_wires(bit_position)
+
+    if max_bitwidth is not None:
+        bitwidth = max_bitwidth
+    else:
+        bitwidth = 2 ** len(bit_position)
+
+    onehot = Const(1, bitwidth=bitwidth)
+
+    return shift_left_logical(onehot, bit_position)
```
fdxmw: Consider adding a comment here on why `shift_left_logical` is needed here - we can't just make a `Const` with a `1` followed by an appropriate number of zeroes because `bit_position` may not be a `Const`, so the number of zeroes may not be known until simulation time

pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
         already_found = already_found | w[i]
 
     return pos
+
+
+def binary_to_one_hot(bit_position, max_bitwidth: int = None) -> WireVector:
+    '''Takes an input representing a bit position and returns a WireVector
+    with that bit position set to 1 and the others to 0.
```
fdxmw, line 1730: Maybe clarify that bit position 0 corresponds to the output's least significant bit

tests/test_helperfuncs.py
```diff
@@ -1814,5 +1814,47 @@ def test_no_ones(self):
         self.assertEqual(sim.inspect('o'), 0)
 
 
+class TestBinaryToOneHot(unittest.TestCase):
+    def setUp(self):
+        pyrtl.reset_working_block()
+
+    def test_simple_binary_to_one_hot(self):
+        i = pyrtl.Input(bitwidth=8, name='i')
```
fdxmw: This would be a little easier to follow if we called this `bit_position` instead of `i` throughout these tests

tests/test_helperfuncs.py
```diff
@@ -1814,5 +1814,47 @@ def test_no_ones(self):
...
+        o = pyrtl.Output(bitwidth=16, name='o')
```
fdxmw: Similarly, consider calling this `one_hot` instead of `o`

pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,37 @@ def one_hot_to_binary(w) -> WireVector:
...
+    '''
+
+    bit_position = as_wires(bit_position)
+
+    if max_bitwidth is not None:
+        bitwidth = max_bitwidth
+    else:
+        bitwidth = 2 ** len(bit_position)
+
+    onehot = Const(1, bitwidth=bitwidth)
```
fdxmw: `onehot` is only used once, so consider inlining it into line 1759 below, like `shift_left_logical(Const(1, bitwidth=bitwidth), bit_position)` . If you want to keep this variable, it should probably be named `one_hot` to be consistent with the function's name (`binary_to_one_hot`)

tests/test_helperfuncs.py
```diff
@@ -1814,5 +1814,47 @@ def test_no_ones(self):
...
+        sim.step({i: 2})
+        self.assertEqual(sim.inspect('o'), 0b0100)
+        sim.step({i: 5})
+        self.assertEqual(sim.inspect('o'), 0b00100000)
+        sim.step({i: 12})
+        self.assertEqual(sim.inspect('o'), 0b0001000000000000)
+        sim.step({i: 15})
+        self.assertEqual(sim.inspect('o'), 0b1000000000000000)
+
+    def test_sufficient_max_bitwidth(self):
```
fdxmw: Consider adding some comments explaining how this test differs from `test_simple_binary_to_one_hot` since they appear to be very similar (this test sets `max_bitwidth` while the previous test does not)

tests/test_helperfuncs.py
```diff
@@ -1814,5 +1814,47 @@ def test_no_ones(self):
...
+        sim.step({i: 6})
+        self.assertEqual(sim.inspect('o'), 0b01000000)
+
+    def test_insufficient_max_bitwidth(self):
+        i = pyrtl.Input(bitwidth=8, name='i')
+        o = pyrtl.Output(bitwidth=16, name='o')
+        o <<= pyrtl.binary_to_one_hot(i, max_bitwidth=4)
+
+        sim = pyrtl.Simulation()
+        sim.step({i: 5})
```
fdxmw: Test the edge case here - `bit_position=4` should be the smallest `bit_position` that returns zero. Maybe test that `bit_position=3` returns `0b1000` while you're at it, since that's the largest `bit_position` that returns nonzero? :) Also consider adding some comments that point out these edge cases

fdxmw (commented):
Thank you for your contribution! Looks good overall, just some comments on documentation and code readability

pyrtl/helperfuncs.py
```diff
@@ -1715,3 +1723,35 @@ def one_hot_to_binary(w) -> WireVector:
...
+def binary_to_one_hot(bit_position, max_bitwidth: int = None) -> WireVector:
+    '''Takes an input representing a bit position and returns a WireVector
+    with that bit position set to 1 and the others to 0.
+
+    :param bit_position: WireVector, WireVector-like object, or something that can be converted
+        into a :py:class:`.Const` (in accordance with the :py:func:`.as_wires()`
+        required input). Example inputs: ``0b10``, ``0b1000``, ``4``.
+    :param max_bitwidth: Optional integer maximum bitwidth for the resulting one-hot WireVector.
+    :return: WireVector with the bit position given by the input set to 1 and all other bits
+    set to 0 (bit position 0 being the least significant bit).
```
fdxmw: I think the alignment is off here, four spaces are needed before `set` so Sphinx understands that this is a continuation of the `:return:` above, and not a new paragraph (see generated docs)

tests/test_helperfuncs.py
```diff
@@ -1814,5 +1814,43 @@ def test_no_ones(self):
...
 
 
+class TestBinaryToOneHot(unittest.TestCase):
+    def setUp(self):
+        pyrtl.reset_working_block()
+
+    def test_simple_binary_to_one_hot(self):
+        bit_position = pyrtl.Input(bitwidth=8, name='bit_position')
+        one_hot = pyrtl.Output(bitwidth=16, name='one_hot')
+        one_hot <<= pyrtl.binary_to_one_hot(bit_position)
```
fdxmw, line 1824: We should also check that `binary_to_one_hot` has the correct bitwidth in both these tests (256 in this example)

fdxmw (commented):
Looks great, just a couple more minor comments, sorry I didn't catch these the first time :(

PR #467, opened 2025-03-06
Use a loop instead of recursion in the find_producer function

gaborszita:
I changed the `find_producer` function in `_ProducerList` from a recursive implementation to an iterative one. As far as I know, recursion is not recommended in production code, especially when a simple iterative approach can achieve the same result.


gaborszita:
made a mistake in my implementation, it is still recursive, closing for now

gaborszita:
fixed

fdxmw (approved):
Thank you for cleaning this up!

PR #470, opened 2025-07-18
Initial implementation of `GateGraph`, an alternative PyRTL logic representation

fdxmw:
`GateGraph` is an alternative PyRTL logic representation that attempts to address several issues with PyRTL's standard `WireVector` and `LogicNet` representation:

1. A `GateGraph` can be directly traversed, without the help of side data structures like the `wire_src_dict` and `wire_sink_dict` returned by `net_connections`.

2. `GateGraph` builds a graph where each node is a `Gate`, and the edges are all references to other `Gates`. By using only one node type, rather than two (`WireVector`, `LogicNet`), it becomes easier to work with the nodes in a `GateGraph`, because every node is the same type, with the same interface.

3. `GateGraph` decouples the user interface and the internal representation. Users directly instantiate `WireVectors` as they build circuits. A user that is only building circuits should never interact with a `GateGraph` or a `Gate`. Users should only interact with `GateGraph` when they need reflection, which typically happens when writing an analysis or optimization pass.

   Another advantage of this decoupling is that `Gate` does not need to support `WireVector`'s quality-of-life features for users, like inferring bitwidth from assignment.

I'm migrating `output_to_verilog` to use this `GateGraph` representation in a separate change.

### DIFF
--- a/docs/blocks.rst
+++ b/docs/blocks.rst
@@ -44,7 +44,7 @@ GateGraphs
 
 .. autoclass:: pyrtl.Gate
     :members:
-    :special-members: __init__, __str__
+    :special-members: __str__
 
 .. autoclass:: pyrtl.GateGraph
     :members:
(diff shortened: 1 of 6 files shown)


fdxmw:
I'm planning to merge this pull request tomorrow, so let me know soon if you have comments.

PR #471, opened 2025-07-23
Update `output_to_verilog` to inline temporary wires, using `GateGraph`:

fdxmw:
- Any Gate with a user-specified name is never inlined.
- Unnamed constant Gates are always inlined.
- Unnamed non-constant Gates are inlined if:
  - The Gate has one user, AND
  - The Gate is not a MemBlock read, AND
  - The Gate is not an arg to a bit-select Gate.

This significantly reduces the amount of generated Verilog code.

Also:

- Update `output_verilog_testbench` to use `GateGraph`.
- Define `GateGraph.__iter__` to make it easier to iterate over all Gates.
- Delete some disabled tests in `test_aes.py`.


fdxmw:
I'm still running more tests on this change. It's potentially controversial, so I'm publishing it in its current state so anyone interested can discuss.

fdxmw:
I've successfully tested this change by exporting [pyrtlnet](https://github.com/UCSBarchlab/pyrtlnet) to Verilog with this change, and simulating the generated Verilog with Verilator.
I'm planning to merge this tomorrow, so let me know soon if you have comments or questions.

PR #472, opened 2025-09-07
Throw an exception if the user tries to render a trace without stepping the simulation

gaborszita:
When the user tries to render a trace without stepping the simulation, PyRTL gives this weird error:

```
gabi@Gabors-MacBook-Pro PyRTL % python3 test.py
Traceback (most recent call last):
  File "/Users/gabi/Documents/PyRTL/test.py", line 8, in <module>
    sim.tracer.render_trace()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/gabi/Documents/PyRTL/pyrtl/simulation.py", line 1787, in render_trace
    self.render_trace_to_text(
    ~~~~~~~~~~~~~~~~~~~~~~~~~^
        trace_list=trace_list,
        ^^^^^^^^^^^^^^^^^^^^^^
    ...<5 lines>...
        segment_size=segment_size,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/Users/gabi/Documents/PyRTL/pyrtl/simulation.py", line 1876, in render_trace_to_text
    current_symbol_len = max(
        len(
    ...<4 lines>...
        for v in trace
    )
ValueError: max() iterable argument is empty
```

`test.py`:

```
import pyrtl

a = pyrtl.Register(name="test_reg", bitwidth=3)
a.next <<= a+1

sim = pyrtl.Simulation()

sim.tracer.render_trace()
```

Output after this PR:

```
gabi@Gabors-MacBook-Pro PyRTL % python3 test.py
Traceback (most recent call last):
  File "/Users/gabi/Documents/PyRTL/test.py", line 8, in <module>
    sim.tracer.render_trace()
    ~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/gabi/Documents/PyRTL/pyrtl/simulation.py", line 1760, in render_trace
    raise PyrtlError("You need to step the simulation at least once to render a trace.")
pyrtl.pyrtlexceptions.PyrtlError: You need to step the simulation at least once to render a trace.
```

I was working on pyrtlnet and it took me around 10 minutes to figure out why I'm getting this error so I decided to open this little PR 😅

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -1755,6 +1755,8 @@ class SimulationTrace:
             is not found in the map, the argument ``repr_func`` will be used instead.
         :param segment_size: Traces are broken in the segments of this number of cycles.
         """
+        if len(self) == 0:
+            raise PyrtlError("You need to step the simulation at least once to render a trace.")
         if repr_per_name is None:
             repr_per_name = {}
         if _currently_in_jupyter_notebook():
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -1756,7 +1756,8 @@ class SimulationTrace:
         :param segment_size: Traces are broken in the segments of this number of cycles.
         """
         if len(self) == 0:
-            raise PyrtlError("You need to step the simulation at least once to render a trace.")
+            msg = "You need to step the simulation at least once to render a trace."
+            raise PyrtlError(msg)
         if repr_per_name is None:
             repr_per_name = {}
         if _currently_in_jupyter_notebook():
--- a/tests/test_compilesim.py
+++ b/tests/test_compilesim.py
@@ -1206,14 +1206,8 @@ class TraceErrorBase(unittest.TestCase):
         r.next <<= r + 1
         sim = self.sim()
         sim.step_multiple(provided_inputs={}, nsteps=10)
-        with self.assertRaises(pyrtl.PyrtlError) as ex:
+        with self.assertRaises(pyrtl.PyrtlError):
             sim.tracer.render_trace()
-        self.assertEqual(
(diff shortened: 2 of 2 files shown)


gaborszita:
Seems a `CompileSim` test is failing, here's why: `render_trace_to_text` throws an exception if we are not tracing any signals and a there is a check for this exception in the `CompileSim` tests. However, `__len__` does the same check, but it just throws the exception with a different message, and since with this change `__len__` is now called before `render_trace_to_text`, `__len__`'s exception is thrown:

image

The two error messages are basically the same thing just described differently, so there is no logic error in the code. There are multiple ways to solve this problem, the simplest is just to copy the error message from `render_trace_to_text` to `__len__`. @fdxmw what do you suggest we do?

fdxmw:
Thank you for improving this!

> The two error messages are basically the same thing just described differently, so there is no logic error in the code. There are multiple ways to solve this problem, the simplest is just to copy the error message from render_trace_to_text to __len__. @fdxmw what do you suggest we do?

This is why I'm generally not a fan of tests that check error messages :) These tests are too sensitive, and often fail when the code's behavior is still correct. I'd be okay with just removing the `assertEqual` that checks the error message; I think it's sufficient to just check that `PyrtlError` was thrown.

Please also run `ruff check` and `ruff format` on your change. Some minor readability improvements will be needed before we can merge it. In the future, please run `tox` to test your changes before you open a pull request; `tox` will also run `ruff` for you. 

In particular, try not to rely on GitHub's pull request actions to run tests for you, because opening a pull request triggers notifications for people watching the repository, which creates unnecessary noise.

PR #475, opened 2025-10-30
Pyrtl floating point library

gaborszita:


pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,342 @@
+import pyrtl
+
+from ._float_utills import FloatUtils
+from ._types import PyrtlFloatConfig, RoundingMode
+
+
+class AddSubHelper:
```
fdxmw: Generally, classes with only `@staticmethod`s probably shouldn't be classes :) It looks like this class can be removed, and all the static methods can be ordinary functions? Same comment for `FloatUtils` and `MultiplicationHelper`

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,342 @@
...
+    @staticmethod
+    def _round(
+        num_mant_bits: int,
+        num_exp_bits: int,
+        raw_result_exponent: pyrtl.WireVector,
+        raw_result_mantissa: pyrtl.WireVector,
+        raw_result_grs: pyrtl.WireVector,
+    ) -> tuple[pyrtl.WireVector, pyrtl.WireVector]:
+        last = raw_result_mantissa[0]
+        guard = raw_result_grs[2]
```
fdxmw: Consider using [`wire_struct`](https://pyrtl.readthedocs.io/en/latest/helpers.html#pyrtl.wire_struct) to simplify these kinds of concat/slice patterns. With a `wire_struct`, you wouldn't need this line, and later lines would instead refer to `raw_result.guard`.

Also see the [`wire_struct` example](https://github.com/UCSBarchlab/PyRTL/blob/development/examples/example1.2-wire-struct.py).

pyrtl/rtllib/pyrtlfloat/_float_utills.py
```diff
@@ -0,0 +1,104 @@
...
+    @staticmethod
+    def get_sign(fp_prop: FPTypeProperties, wire: pyrtl.WireVector) -> pyrtl.WireVector:
+        return wire[fp_prop.num_mantissa_bits + fp_prop.num_exponent_bits]
+
+    @staticmethod
+    def get_exponent(
+        fp_prop: FPTypeProperties, wire: pyrtl.WireVector
+    ) -> pyrtl.WireVector:
+        return wire[
+            fp_prop.num_mantissa_bits : fp_prop.num_mantissa_bits
```
fdxmw: Similarly, [`wire_struct`](https://pyrtl.readthedocs.io/en/latest/helpers.html#pyrtl.wire_struct) can remove the need for these tricky bit offset calculations

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,342 @@
+import pyrtl
+
+from ._float_utills import FloatUtils
+from ._types import PyrtlFloatConfig, RoundingMode
+
+
+class AddSubHelper:
+    @staticmethod
+    def add(
```
fdxmw: We'll need docstrings for all user-facing classes, methods, and functions. Docstrings for internal stuff would also be great; for example as someone new to this code it would really help to have an idea of what `make_denormals_zero` does conceptually.

pyrtl/rtllib/pyrtlfloat/_float_utills.py
```diff
@@ -0,0 +1,104 @@
...
+        return wire[: fp_prop.num_mantissa_bits]
+
+    @staticmethod
+    def is_zero(fp_prop: FPTypeProperties, wire: pyrtl.WireVector) -> pyrtl.WireVector:
+        return (FloatUtils.get_mantissa(fp_prop, wire) == 0) & (
+            FloatUtils.get_exponent(fp_prop, wire) == 0
+        )
+
+    @staticmethod
+    def is_inf(fp_prop: FPTypeProperties, wire: pyrtl.WireVector) -> pyrtl.WireVector:
```
fdxmw: It would really help to have some comments with brief definitions and references for important floating point concepts (NaN, inf, denormalized (how are these different?), guard, round, sticky, ...). 

I'm not expecting you to teach me everything in comments, but it would really help to know what I should read if I want to understand how this works :)

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,161 @@
...
+        config: PyrtlFloatConfig,
+        operand_a: pyrtl.WireVector,
+        operand_b: pyrtl.WireVector,
+    ) -> pyrtl.WireVector:
+        fp_type_props = config.fp_type_properties
+        rounding_mode = config.rounding_mode
+        num_exp_bits = fp_type_props.num_exponent_bits
+        num_mant_bits = fp_type_props.num_mantissa_bits
+
+        operand_a_daz = FloatUtils.make_denormals_zero(fp_type_props, operand_a)
```
fdxmw: There are a lot of paired lines of code in this function, where the two lines in the pair are nearly identical, like this line and line 20 below.

It seems clearer to put the operands in a tuple, and put the operation's output in a tuple, so we don't have to repeat ourselves, or worry about the paired lines going out of sync. Something like

```
operands = (operand_a, operand_b)
operands_daz = (
    FloatUtils.make_denormals_zero(fp_type_props, operand) 
    for operand in operands
)
```

We could try to improve this further with `dataclass` so we can say `operand.a` instead of `operand[0]`, but that seems marginal to me since the names `a` and `b` were arbitrary in the first place (could have also named them `left` and `right` :)

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+            exponent_min_value = (
+                sum_exponent_min_value - need_to_normalize - exponent_incremented
+            )
+        else:
+            exponent_max_value = sum_exponent_max_value - need_to_normalize
+            exponent_min_value = sum_exponent_min_value - need_to_normalize
+
+        if rounding_mode == RoundingMode.RNE:
+            raw_result_exponent = rounded_product_exponent[0:num_exp_bits]
```
fdxmw: I think you can omit the `0` here, also on line 119 below. Doing so would be more consistent with your code above, line 63 for example.

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,161 @@
...
+                        exponent_incremented |= 0
+                with pyrtl.otherwise:
+                    rounded_product_mantissa |= normalized_product_mantissa
+                    rounded_product_exponent |= normalized_product_exponent
+                    exponent_incremented |= 0
+
+        result_exponent = pyrtl.WireVector(bitwidth=num_exp_bits)
+        result_mantissa = pyrtl.WireVector(bitwidth=num_mant_bits)
+
+        operand_a_nan = FloatUtils.is_NaN(fp_type_props, operand_a_daz)
```
fdxmw: Capitalization of `NaN` is inconsistent here, it's `nan` on the left hand side and `NaN` on the right hand side. I'd rename `is_NaN` to `is_nan` since we pretty consistently use lowercase with underscores for method and function names.

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,30 @@
...
+        b_floatwv <<= b
+        FloatOperations.default_rounding_mode = RoundingMode.RNE
+        result_add = pyrtl.Output(name="result_add")
+        result_add <<= a_floatwv + b_floatwv
+        result_sub = pyrtl.Output(name="result_sub")
+        result_sub <<= a_floatwv - b_floatwv
+        self.sim = pyrtl.Simulation()
+
+    def test_multiplication_simple(self):
+        self.sim.step({"a": 0b0100001000000000, "b": 0b0100010100000000})
```
fdxmw: We'll need a better story for how this is tested :) A few things to consider:

Every PyRTL developer will need to run these tests for the indefinite future, so we really don't want the tests to run for too long. I'd say all continuously-run floating point tests should complete in a couple seconds or less. These tests can't be comprehensive, but we should still try to get as much value as we can out of them.

Try to think about the most interesting inputs (zeroes, inf, nan, ...), and which combinations of inputs and operations are worth testing. I would avoid randomly generated tests because they are unlikely to cover all the interesting combinations with a reasonable number of test cases.

Think about how someone besides yourself would debug one of these test failures. It would help a lot to know what cases a particular test exercises (`# Test addition with inf`, etc). It also helps to know if the unexpected output is slightly wrong, or complete nonsense, which is pretty hard to tell from a raw bit pattern (see next point :)

Consider adding some helper functions that can convert `float` to and from these bit patterns. That would make it easier for someone to understand that the test checks that 1.0 + 2.0 == 3.0, for example.

pyrtl/rtllib/pyrtlfloat/floatoperations.py
```diff
@@ -2,14 +2,14 @@
...
-from ._types import PyrtlFloatConfig, RoundingMode
+from ._types import FloatingPointType, PyrtlFloatConfig, RoundingMode
 
 
 class FloatOperations:
     default_rounding_mode = RoundingMode.RNE
 
     @staticmethod
-    def multiply(
+    def mul(
```
fdxmw, line 39: This doesn't seem very useful, since it just passes all its arguments through to `MultiplicationHelper.multiply`? Seems better to point users directly to `MultiplicationHelper.multiply`. Same comments for the other methods below.

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,342 @@
...
+        config: PyrtlFloatConfig,
+        operand_a: pyrtl.WireVector,
+        operand_b: pyrtl.WireVector,
+    ) -> pyrtl.WireVector:
+        fp_type_props = config.fp_type_properties
```
fdxmw: All user-facing methods/functions should check (somewhere, not necessarily here) that the operand bitwidths are consistent with `config`

fdxmw (commented):
Thank you for this contribution! Here are some initial comments

tests/rtllib/pyrtlfloat/test_add_sub.py
gaborszita: Yep, testing needs more work, I'll add more rigorous tests and test edge cases. That's why I marked it as a draft PR because it is not finished yet.

pyrtl/rtllib/pyrtlfloat/floatoperations.py
gaborszita: The reason I have this `FloatOperations` class and separate `MultiplicationHelper` and `AddSubHelper` classes is because the user is supposed to use `FloatOperations`, `MultiplicationHelper` and `AddSubHelper` are internal helpers and the user is not supposed to use them. I created these helpers so I can separate the multiplication and addition/subtraction logic into separate files.

fdxmw: I think that makes sense, but I'd still try to refactor, it would be nice to remove these functions that are just pass-through wrappers for another function. Maybe:

1. `AddSubHelper.add` could be an ordinary function instead of a `@staticmethod`
2. `_add_sub.py` could be renamed to `add_sub.py`
3. Non-public functions in `add_sub.py` could have an underscore prefix
4. `__init__.py` could import `add` from `add_sub`

I think that would still make it clear which parts of the interface are meant for public use, while removing a layer of indirection?

This kind of indirection tends to be annoying when debugging problems, because you have to jump through another hoop to find the code you're looking for. ["Our princess is in another castle!"](https://en.wikipedia.org/wiki/Our_princess_is_in_another_castle!)

pyrtl/rtllib/pyrtlfloat/_float_utills.py
gaborszita: I was trying to do this, but the problem is that the number of mantissa and exponent bits differs based on `fp_prop`, so we would need to set the bitwidths of the slices in the `wire_struct` dynamically at runtime. Is it possible to do this with `wire_struct`?

fdxmw: It is possible with some trickery, but it's probably overkill in this case, since you only need to select between a few well-known bit layouts (the user can't define an arbitrary custom floating point format). I think you could define a set of `wire_struct`s that share the same interface, something like:

```python
@pyrtl.wire_struct
class Float16
    sign: 1
    exponent: 5
    fraction: 10

@pyrtl.wire_struct
class Float32
    sign: 1
    exponent: 8
    fraction: 23
```

So the idea is that these objects share the same interface, so it's easier to write code that works with any of these types. For example, if you want to compare sign bits, you'd write `a.sign == b.sign`, and that would work regardless of whether `a` or `b` are `Float16` or `Float32`. If you care about types (I encourage you to! :), these are [`Union`](https://docs.python.org/3/library/typing.html#typing.Union)s like `Float16 | Float32`.

In case it's useful, here's the trickery, you can define a function that returns a dynamically-defined class:

```python
import pyrtl

def define_my_struct(a_bits: int, b_bits: int):
    @pyrtl.wire_struct
    class MyInternalClass:
        a: a_bits
        b: b_bits

    return MyInternalClass

MyStruct = define_my_struct(a_bits=4, b_bits=8)

my_struct_instance = MyStruct(a=1, b=2)

print("a bitwidth", my_struct_instance.a.bitwidth)
print("b bitwidth", my_struct_instance.b.bitwidth)
print("total bitwidth", my_struct_instance.bitwidth)
```

```shell
$ uv run ...
a bitwidth 4
b bitwidth 8
total bitwidth 12
```

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+
+    # Align mantissas by shifting the smaller one to match the larger's exponent.
+    smaller_mantissa_shift_amount = larger_operand_exponent - smaller_operand_exponent
+    smaller_mantissa_shifted = pyrtl.shift_right_logical(
+        smaller_operand_mantissa, smaller_mantissa_shift_amount
+    )
+
+    # RNE rounding uses the guard, round, and sticky bits.
+    # When shifting the smaller mantissa to the right, some bits are shifted out.
+    # The first bit shifted out becomes the guard bit, the second becomes the round bit,
```
fdxmw: The words `first` and `second` here are a bit confusing because I can interpret it the other way, for example if I shift the bits one at a time, the `first` bit that gets shifted out is the least significant bit (`smaller_operand_mantissa[0]`), regardless of the `shift_amount`. 

Maybe the `first` bit could be something like "the most significant bit to the right of the mantissa, after shifting"?

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+    # and any remaining bits are ORed together to form the sticky bit.
+    # https://drilian.com/posts/2023.01.10-floating-point-numbers-and-rounding/
+    grs = pyrtl.WireVector(bitwidth=3)  # guard, round, sticky bits
+    with pyrtl.conditional_assignment:
+        # If the smaller mantissa is shifted by 2 or more, the first two bits
+        # shifted out are the guard and round bits, and the sticky bit is
+        # the OR of all remaining bits.
+        with smaller_mantissa_shift_amount >= 2:
+            guard_and_round = pyrtl.shift_right_logical(
```
fdxmw, line 162: This is fine for now, but keep in mind that `shift_right_logical` creates a barrel shifter, which is pretty expensive, and we just made one on L75 above (and we'll make another on L95 below :). It should be possible to use one barrel shifter instead of three, by appending zero-valued GRS bits to the `smaller_operand_mantissa` before we shift it. That should work for the guard and round bits, but properly tracking the sticky bit would likely require changes to the barrel shifter.

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+    smaller_operand_exponent = get_exponent(fp_type_props, operand_smaller)
+    larger_operand_exponent = get_exponent(fp_type_props, operand_larger)
+    smaller_operand_mantissa = pyrtl.concat(
+        pyrtl.Const(1), get_mantissa(fp_type_props, operand_smaller)
+    )
+    larger_operand_mantissa = pyrtl.concat(
+        pyrtl.Const(1), get_mantissa(fp_type_props, operand_larger)
+    )
+
+    # Align mantissas by shifting the smaller one to match the larger's exponent.
```
fdxmw, line 141: I think an important insight here is:
1. Shifting the mantissa right by one divides the value by two.
2. Adding one to the exponent multiplies the value by two. 
3. If we simultaneously do (1) and (2), we don't change the value, but it helps match the operand's exponents for easier math.

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+
+    # Perform addition of operands.
+    sum_exponent, sum_mantissa, sum_grs, sum_carry = _add_operands(
+        larger_operand_exponent,
+        smaller_mantissa_shifted_grs,
+        larger_mantissa_extended,
+    )
+
+    # Perform subtraction of operands.
+    sub_exponent, sub_mantissa, sub_grs, num_leading_zeros = _sub_operands(
```
fdxmw: Naming seems a little inconsistent here, above `_add_operands` returns `sum_exponent`, while here `_sub_operands` returns `sub_exponent`. Seems like this should be called `difference_exponent` (addition results in a sum, subtraction results in a difference)?

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+                - rounding_exponent_incremented
+            )
+        else:
+            larger_exponent_max_value = initial_larger_exponent_max_value - sum_carry
+        # Checks if an addition was performed and the result overflowed.
+        with (smaller_operand_sign == larger_operand_sign) & (
+            larger_operand_exponent > larger_exponent_max_value
+        ):
+            final_result_sign |= larger_operand_sign
+            if rounding_mode == RoundingMode.RNE:
```
fdxmw, line 384: It's pretty counterintuitive that `rounding_mode` determines whether you get `inf` or `largest_finite_number` on overflow. A reference would be really useful here

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+        # We check for underflow by computing the min value of the larger
+        # operand's exponent. As with overflow, this value can vary depending
+        # on the operands. We subtract the number of leading zeros from the
+        # larger exponent to obtain the subtraction exponent. Additionally,
+        # if rounding causes the exponent to increment, we need to account
+        # for that. Therefore, we add the number of leading zeros and
+        # subtract the rounding increment from the absolute minimum exponent,
+        # which is one greater than the all-0s exponent (reserved for
+        # zero and denormals).
+        initial_larger_exponent_min_value = pyrtl.Const(1)
```
fdxmw: Seems like this initialization should be `= None`, since it always gets overwritten in the `if` statement below? (calling `pyrtl.Const(1)` creates a `Const` and adds it to the `Block`, but that `Const` is never used)

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+        with pyrtl.otherwise:
+            raw_result_exponent |= sub_exponent
+            raw_result_mantissa |= sub_mantissa
+            if rounding_mode == RoundingMode.RNE:
+                raw_result_grs |= sub_grs
+
+    # Round the result if using RNE rounding mode.
+    if rounding_mode == RoundingMode.RNE:
+        (
+            raw_result_rounded_exponent,
```
fdxmw: It doesn't seem useful to give the rounded results new names (`raw_result_rounded_exponent`, etc), since we don't seem to use the not-rounded results (`raw_result_exponent`, etc). If this instead updated `raw_result_exponent` and friends, I think we could remove the conditional on L287 below, since we can always assign `final_result_exponent |= raw_result_exponent` ?

Also this function is quite long and there are a lot of variables, so it would really help readers to clarify variable lifetimes (which variables are still in use), and how groups of variables are related. Some ways to do this include:
1. Avoid creating new variables when it makes sense to reuse an existing one (this suggestion)
2. Try to define fewer variables with more internal structure, rather than defining a flat layer of top-level variables (`raw_result.exponent` rather than `raw_result_exponent`) (an earlier suggestion)
3. Factor out more functions, which clearly defines the inputs and outputs for each chunk of code, and creates more local variable scopes
4. Explicitly `del` variables that aren't needed anymore

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+            larger_exponent_min_value = (
+                initial_larger_exponent_min_value + num_leading_zeros
+            )
+        # Checks if a subtraction was performed and the result underflowed.
+        with (smaller_operand_sign != larger_operand_sign) & (
+            larger_operand_exponent < larger_exponent_min_value
+        ):
+            final_result_sign |= larger_operand_sign
+            make_zero(final_result_exponent, final_result_mantissa)
+        with pyrtl.otherwise:
```
fdxmw, line 397: Could use a comment like "Otherwise no special cases apply: this is the common case" :)

fdxmw (commented):
Sorry for the slow review! This code is making a lot more sense to me now, and I think it looks pretty good! I've only reviewed the main addition and subtraction code for now, I'll take a look at multiplication tomorrow. Here are some comments in the meantime.

Also I still stand by my earlier comments about trying to remove layers of indirection that don't do much work (`FloatOperations`, `_BaseTypedFloatOperations`) and trying to use `WireStruct` instead of calculating bit offsets :)

pyrtl/rtllib/pyrtlfloat/_float_utills.py
fdxmw, line 1: This should be called `_float_utils.py`? (only one L in utilities :)

pyrtl/rtllib/pyrtlfloat/_float_utills.py
```diff
@@ -0,0 +1,173 @@
...
+    return (get_mantissa(fp_prop, wire) == 0) & (
+        get_exponent(fp_prop, wire) == (1 << fp_prop.num_exponent_bits) - 1
+    )
+
+
+def is_denormalized(
+    fp_prop: FPTypeProperties, wire: pyrtl.WireVector
+) -> pyrtl.WireVector:
+    """
+    Returns whether the floating point number is denormalized.
```
fdxmw: A reference that defines denormalized numbers would be very helpful. https://en.wikipedia.org/wiki/Subnormal_number seems good, but maybe you've found something better?

pyrtl/rtllib/pyrtlfloat/_float_utills.py
```diff
@@ -0,0 +1,173 @@
...
+                get_sign(fp_prop, wire),
+                get_exponent(fp_prop, wire),
+                pyrtl.Const(0, bitwidth=fp_prop.num_mantissa_bits),
+            )
+        with pyrtl.otherwise:
+            out |= wire
+    return out
+
+
+def make_inf(
```
fdxmw: It's a little confusing that `make_denormals_zero` returns a `WireVector` while `make_inf` conditionally assigns its argument `WireVectors`. Consistency is clearer, so try to be consistent :)

I'd recommend just returning `WireVectors` in all these methods because the implementation is much easier to understand. Very few people know what happens when you do a `|=` assignment outside of a `conditional_assignment` block.

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    :return: The result of the multiplication as a WireVector.
+    """
+    fp_type_props = config.fp_type_properties
+    rounding_mode = config.rounding_mode
+    num_exp_bits = fp_type_props.num_exponent_bits
+    num_mant_bits = fp_type_props.num_mantissa_bits
+
+    # Denormalized numbers are not supported, so we flush them to zero.
+    operands = (operand_a, operand_b)
+    operands_daz = tuple(make_denormals_zero(fp_type_props, op) for op in operands)
```
fdxmw: I left a similar comment on `_add_sub.py`, but I'd redefine `operands` here rather than defining a new `operands_daz` since we don't seem to use `operands` anymore after this point. 

Similarly, this is a big function, so think about how someone new to this code will read it, and look for ways to help them navigate and understand what's important.

Don't forget that you have the Curse of Knowledge: you wrote all this code, so it already makes perfect sense to you :) To really test your code's readability, you have to show it to other people.

pyrtl/rtllib/pyrtlfloat/_add_sub.py
```diff
@@ -0,0 +1,447 @@
...
+    :return: The result of the addition as a WireVector.
+    """
+    fp_type_props = config.fp_type_properties
+    rounding_mode = config.rounding_mode
+    num_exp_bits = fp_type_props.num_exponent_bits
+    num_mant_bits = fp_type_props.num_mantissa_bits
+    total_bits = num_exp_bits + num_mant_bits + 1
+
+    # Denormalized numbers are not supported, so we flush them to zero.
+    operand_a_daz = make_denormals_zero(fp_type_props, operand_a)
```
fdxmw: Consider grouping the operands into a tuple, for consistency with the multiplication implementation, and removing redundant code. Maybe it should be a sorted tuple, so the smaller operand is `operands[0]` and we wouldn't need more names to track smaller/larger?

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # Extract the sign and exponent of both operands.
+    signs = tuple(get_sign(fp_type_props, op) for op in operands_daz)
+    exponents = tuple(get_exponent(fp_type_props, op) for op in operands_daz)
+
+    result_sign = signs[0] ^ signs[1]
+
+    # IEEE-754 floating point numbers have a bias:
+    # https://en.wikipedia.org/wiki/Exponent_bias
+    # real_exponent = stored_exponent - bias, so stored_exponent = real + bias
```
fdxmw: Maybe an easier way to think about this (for me, at least :)

`stored_exponent = real + bias`, so when you add two `stored_exponent`s together, you get 

`(real0 + bias) + (real1 + bias) = real0 + real1 + 2 * bias`

Which is a double-biased result. We can correct that to a single-bias result by subtracting off `bias` to produce `real0 + real1 + bias`

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    mantissas = tuple(
+        pyrtl.concat(pyrtl.Const(1), get_mantissa(fp_type_props, op))
+        for op in operands_daz
+    )
+    product_mantissa = mantissas[0] * mantissas[1]
+
+    normalized_product_exponent = pyrtl.WireVector(bitwidth=num_exp_bits + 1)
+    normalized_product_mantissa = pyrtl.WireVector(bitwidth=num_mant_bits)
+
+    # We need to normalize (shift right) if the leading bit is 1.
```
fdxmw: I think it's worth explaining why normalization requires at most one right shift. It took me a few minutes to figure out, even with the reference below which doesn't really spell it out. Or maybe we can find a better reference?

IIUC, we're multiplying two equal-bitwidth numbers that both have their most significant bit set to 1, with binary points just to the right of the MSB. So the inputs both look like `1.<something>` in binary.

The product's binary point will be just after its second-most significant bit, so the output will look like `ab.cdef...` in binary, where each letter is one bit.

A properly normalized output must have the form `1.<something>` so if `a` (the output's MSB) is `1`, we must normalize by right shifting one position and increasing the exponent by 1. If `a` is 0 we don't need to do anything.

We'll never need to shift right more than one position because the biggest output we can produce is `1.111... * 1.111... = 11.111...`. So it's not possible to have more than two bits to the left of the binary point. 

We'll never need to shift left because you can't multiply two `1.<something>`s together and end up with a number less than 1 (`ab != 0`; adding an `rtl_assert` for this would make this clearer).

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    # https://numeral-systems.com/ieee-754-multiply/
+    need_to_normalize = product_mantissa[-1]
+
+    if rounding_mode == RoundingMode.RNE:
+        guard = pyrtl.WireVector(bitwidth=1)
+        sticky = pyrtl.WireVector(bitwidth=1)
+        last = pyrtl.WireVector(bitwidth=1)  # Last bit of the mantissa before rounding.
+
+    # Assign the normalized mantissa, exponent, guard, sticky, and last bits
+    # based on whether normalization is needed.
```
fdxmw: Add a reference to https://drilian.com/posts/2023.01.10-floating-point-numbers-and-rounding/ for `guard` and `sticky` bits. It'd also help to explain why we don't need a `round` bit for multiplication (seems related to `ab != 0` above but I'm not sure)

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    # We need to normalize (shift right) if the leading bit is 1.
+    # https://numeral-systems.com/ieee-754-multiply/
+    need_to_normalize = product_mantissa[-1]
+
+    if rounding_mode == RoundingMode.RNE:
+        guard = pyrtl.WireVector(bitwidth=1)
+        sticky = pyrtl.WireVector(bitwidth=1)
+        last = pyrtl.WireVector(bitwidth=1)  # Last bit of the mantissa before rounding.
+
+    # Assign the normalized mantissa, exponent, guard, sticky, and last bits
```
fdxmw: It seems clearer to always mention `guard`, `sticky` and `last` in the order they appear in the un-normalized mantissa, left to right (so `last, guard, and sticky`). When possible, we should also work with them in this order.

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    # based on whether normalization is needed.
+    with pyrtl.conditional_assignment:
+        with need_to_normalize:
+            normalized_product_mantissa |= product_mantissa[-num_mant_bits - 1 :]
```
fdxmw: I think this logic can be simplified by only aligning the `product_mantissa` binary points under this `conditional_assignment`, so all the other bits can be determined unconditionally outside the `conditional_assignment`. Something like this (untested! :)

```python
# Align the mantissa's binary point into the standard 1.<something> 
# format. 
with pyrtl.conditional_assignment:
    with need_to_normalize:
        # product_mantissa's MSB is 1 so we just reinterpret the 1b.cdef... 
        # format as 1.bcdef...
        aligned_mantissa |= product_mantissa
        normalized_product_exponent |= product_exponent + 1
    with pyrtl.otherwise:
        # Output is 01.<something> so shift it left to drop the leading 
        # zero and put it in the standard 1.<something> format.
        aligned_mantissa |= pyrtl.Concat(product_mantissa[:-1], pyrtl.Const(0, bitwidth=1))
        normalized_product_exponent |= product_exponent

# Extract the parts we need from the aligned_mantissa.
normalized_product_mantissa |= aligned_mantissa[-num_mant_bits - 1 :]
last = aligned_mantissa[-num_mant_bits - 1]
guard = aligned_mantissa[-num_mant_bits - 2]
sticky = aligned_mantissa[: -num_mant_bits - 2] != 0
```

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+        rounded_product_mantissa = pyrtl.WireVector(bitwidth=num_mant_bits)
+        rounded_product_exponent = pyrtl.WireVector(bitwidth=num_exp_bits + 1)
+        # Whether exponent was incremented due to rounding (for overflow check).
+        exponent_incremented = pyrtl.WireVector(bitwidth=1)
+        # If guard bit is not set, number is closer to smaller value: no round.
+        # If guard and sticky are set, round up.
+        # If guard is set but sticky is not, value is exactly halfway.
+        # Following round-to-nearest ties-to-even, round up if last bit is 1.
+        round_up = guard & (last | sticky)
+        with pyrtl.conditional_assignment:
```
fdxmw: This looks identical to the `conditional_assignment` in `_round`, can we factor out the repeated code?

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+                exponent_incremented |= 0
+
+    result_exponent = pyrtl.WireVector(bitwidth=num_exp_bits)
+    result_mantissa = pyrtl.WireVector(bitwidth=num_mant_bits)
+
+    # Check whether operands are special: NaN, infinity, zero, or denormalized.
+    operand_nans = tuple(is_nan(fp_type_props, op) for op in operands_daz)
+    operand_infs = tuple(is_inf(fp_type_props, op) for op in operands_daz)
+    operand_zeros = tuple(is_zero(fp_type_props, op) for op in operands_daz)
+    operand_denorms = tuple(is_denormalized(fp_type_props, op) for op in operands_daz)
```
fdxmw: Nit: Try to be consistent with your names, this breaks the pattern established on lines 125-127 (`zeros` : `zero` :: `denorms` : `denormalized`)

pyrtl/rtllib/pyrtlfloat/_float_utills.py
```diff
@@ -0,0 +1,173 @@
...
+    Returns the mantissa bits of floating point number.
+
+    :param fp_prop: Floating point type properties.
+    :param wire: WireVector holding the floating point number.
+    :return: WireVector holding the mantissa bits.
+    """
+    return wire[: fp_prop.num_mantissa_bits]
+
+
+def is_zero(fp_prop: FPTypeProperties, wire: pyrtl.WireVector) -> pyrtl.WireVector:
```
fdxmw: Seems like we always run all these `is_*` checks at the same time, so it seems clearer to define them in one `check_kinds` function that does all the checks and figures out what kind of number we have. The function can return a `wire_struct` of 4 named bits to reduce the number of top-level variables, something like

```python
smaller_kinds = check_kinds(smaller_operand)
...
with smaller_kinds.is_nan:
...
```

I think the kinds are mutually exclusive? If that's right, it could make sense to encode them into 2 bits instead of using 4 one-hot bits.

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+
+    # We check for overflow and underflow by computing max and min exponent
+    # values of the sum of operands before rounding and normalization.
```
fdxmw: Should "sum of operands" say "sum of operands' exponents" (we're adding the exponents to multiply the operands) ?

pyrtl/rtllib/pyrtlfloat/_multiplication.py
```diff
@@ -0,0 +1,197 @@
...
+    # may further increase the exponent. Therefore, we subtract these
+    # potential increments from the absolute maximum exponent, which is one
+    # less than the all-1s exponent (reserved for inf/NaN) plus bias.
+    # Similarly, we subtract these increments from the absolute minimum
+    # exponent, which is 1 plus the exponent bias.
+    sum_exponent_max_value = pyrtl.Const(2**num_exp_bits - 2 + exponent_bias)
+    sum_exponent_min_value = pyrtl.Const(1 + exponent_bias)
+    if rounding_mode == RoundingMode.RNE:
+        exponent_max_value = (
+            sum_exponent_max_value - need_to_normalize - exponent_incremented
```
fdxmw: This works, but think about the generated logic: each `-` in this line instantiates a full adder, but one of them is only used to subtract a 1-bit number from a constant, and the other just subtracts another 1-bit number from the first subtraction's result.

So we're really just selecting between:
1. Constant
2. Constant - 1
3. Constant - 2 

which shouldn't require instantiating any adders at all. If you refactor this, definitely add a comment explaining why it's not written as two subtractions, so someone doesn't try to "fix" it later :)

Similar comments below, also for the similar code in `add()`.

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+FLOAT16_ONE_POINT_FIVE = 0x3E00  # 1.5
+FLOAT16_ONE_POINT_TWOFIVE = 0x3D00  # 1.25
+FLOAT16_LARGEST_NORMAL = 0x7BFF  # Largest normal number (~65504)
+FLOAT16_SMALLEST_NORMAL = 0x0400  # Smallest normal number (2^-14)
+FLOAT16_DENORMAL = 0x0001  # Smallest denormal
+
+
+def float16_parts(sign, exp, mant):
+    """Construct Float16 from sign, exponent, and mantissa."""
+    return (sign << 15) | (exp << 10) | mant
```
fdxmw: An `assert` that each part fits in its bitwidth would help prevent future test writers from shooting themselves in the foot :)

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+
+
+def decode_float16(bits):
+    """Decode Float16 bits to sign, exponent, and mantissa."""
+    return (bits >> 15) & 1, (bits >> 10) & 0x1F, bits & 0x3FF
```
fdxmw: Similarly, `assert` that `bits` fits in 16 bits

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+        self.sim.step({"a": FLOAT16_ONE, "b": FLOAT16_NEG_ONE})
+        self.assertEqual(self.sim.inspect("result_rne"), FLOAT16_POS_ZERO)
+        self.assertEqual(self.sim.inspect("result_rtz"), FLOAT16_POS_ZERO)
+
+
+class TestAdditionRounding(unittest.TestCase):
+    """Tests for rounding in addition (RNE and RTZ)."""
+
+    def setUp(self):
+        pyrtl.reset_working_block()
```
fdxmw: All these tests seem to use the same hardware, so it would help to factor this logic out to make it obvious that we're still using the same hardware, or to clearly point out any differences. Trying to visually identify any differences as a reader is not easy :) Some ideas:
1. Just define all the tests that use the same hardware in the same class. You can add big section comments like
   ```python
   ############################
   # Additional rounding tests.
   ```
   to group the different kinds of tests for people reading the code. I recommend this approach.
2. Factor the hardware definition out into a function.
3. Move the hardware definition to a base class.

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+        self.assertTrue(is_nan(self.sim.inspect("result_rtz")))
+
+    def test_add_nan_propagates(self):
+        """Test x + NaN = NaN"""
+        self.sim.step({"a": FLOAT16_ONE, "b": FLOAT16_NAN})
+        self.assertTrue(is_nan(self.sim.inspect("result_rne")))
+        self.assertTrue(is_nan(self.sim.inspect("result_rtz")))
+
+    def test_add_denormal_flushed_to_zero(self):
+        """Test that denormal inputs are flushed to zero."""
```
fdxmw: Same nit about consistency, I've seen at least `denormal`, `denorm`, and `denormalized` in this PR. Try to pick one and use it consistently :)

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+        result_rne <<= Float16Operations.add(self.a, self.b)
+        FloatOperations.default_rounding_mode = RoundingMode.RTZ
+        result_rtz = pyrtl.Output(name="result_rtz")
+        result_rtz <<= Float16Operations.add(self.a, self.b)
+        self.sim = pyrtl.Simulation()
+
+    def test_add_one_plus_two(self):
+        """Test 1.0 + 2.0 = 3.0"""
+        self.sim.step({"a": FLOAT16_ONE, "b": FLOAT16_TWO})
+        self.assertEqual(self.sim.inspect("result_rne"), FLOAT16_THREE)
```
fdxmw: I'm sure you've run into this, but think about how someone would debug one of these `assertEqual` failures. I think you'd get a message that says 
```
{big decimal number} is not equal to {other big decimal number}
```
which is not very helpful.

Most likely, the first thing you'd do to debug a failure is to call `decode_float16` to split out the sign, exponent, and mantissa on the unexpected value. Since this is the most likely first debug step, it'd really help the next person that works on this (which could be you! :) to integrate `decode_float16` into these checks.

Maybe we can define a `assertFloat16Equal` helper that internally calls `decode_float16` and calls `assertEqual` on the `(sign, exponent, mantissa)` tuple?

tests/rtllib/pyrtlfloat/test_add_sub.py
```diff
@@ -0,0 +1,525 @@
...
+
+import pyrtl
+from pyrtl.rtllib.pyrtlfloat import Float16Operations, FloatOperations, RoundingMode
+
+# IEEE 754 Float16 special values
+FLOAT16_POS_ZERO = 0x0000
+FLOAT16_NEG_ZERO = 0x8000
+FLOAT16_POS_INF = 0x7C00
+FLOAT16_NEG_INF = 0xFC00
+FLOAT16_NAN = 0x7E00  # Quiet NaN with mantissa bit set
```
fdxmw: Add links to https://en.wikipedia.org/wiki/NaN#Quiet_NaN and https://en.wikipedia.org/wiki/NaN#Encoding ?

gaborszita:
Seems that we cannot import from helper files when running tests... I didn't catch this error because when locally I ran `python3 -m tests.rtllib.pyrtlfloat.test_add_sub` it worked, seems it only fails when running with `uv`. What solution would you recommend? I  just created this helper file to avoid duplicating code.

gaborszita:
Also sorry for the late fix, I just got very busy the past month and I'm still catching up on overdue things.

fdxmw:
> Seems that we cannot import from helper files when running tests... I didn't catch this error because when locally I ran `python3 -m tests.rtllib.pyrtlfloat.test_add_sub` it worked, seems it only fails when running with `uv`. What solution would you recommend? I just created this helper file to avoid duplicating code.

I highly recommend running tests with `uv`, it ensures that your tests run in the same environment as the automated testing. I use commands like this:

```
$ uv run tests/rtllib/test_adders.py TestAdders.test_ripple_1 -v
test_ripple_1 (__main__.TestAdders.test_ripple_1) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.007s

OK
```

gaborszita:
I tried running it with `uv`, and got the same error as in the github workflow run. It seems the way the project is set up it is not possible for a test file to import from another file in the `tests` directory.

I have three potential solutions:

1. Duplicate the shared code to both test files.
2. Combine the two test files into one file.
3. Move the helper test file to somewhere in the pyrtl directory so the test files can properly import it. The problem with this (in my opinion) is that it will mess up the file organization (files related to tests should go in the tests directory).

Which solution would you recommend? Personally I think #2 is the best.

fdxmw:
> I tried running it with `uv`, and got the same error as in the github workflow run. It seems the way the project is set up it is not possible for a test file to import from another file in the `tests` directory.
> 
> I have three potential solutions:
> 
>     1. Duplicate the shared code to both test files.
> 
>     2. Combine the two test files into one file.
> 
>     3. Move the helper test file to somewhere in the pyrtl directory so the test files can properly import it. The problem with this (in my opinion) is that it will mess up the file organization (files related to tests should go in the tests directory).
> 
> 
> Which solution would you recommend? Personally I think #2 is the best.

Hmm I agree it's annoying that we can't keep test utilities in the test directory. I tried searching around and the closest thing I found to a solution was `pytest`'s `pythonpath` option, but that only works when the tests are run through `pytest`, which is not great. I'd really like the tests to just work when they're run directly through the interpreter.

Combining the tests (2) sounds good to me. I'd also be okay with (3), which I agree is not great, but also doesn't seem too bad. There's also already some precedent for (3) in https://github.com/UCSBarchlab/PyRTL/blob/development/pyrtl/rtllib/testingutils.py, which is also due for a cleanup.

And sorry I misunderstood your earlier comment! I thought you were just asking about how to run the tests locally :(

fdxmw (approved):
Thank you for addressing all my comments! I think this PR looks good and we can merge it after addressing the test failure. 

I have some more refactoring ideas, but I can experiment with that on my own after merging this PR.

PR #476, opened 2025-10-31
Migrate PyRTL to `uv`

fdxmw:
Migrating to `uv` has several advantages:

* The new workflow is significantly faster. On my computer: 
  ```
  $ # New workflow
  $ uv run just tests
  ...
  real    0m11.830s
  user    1m31.411s
  sys     0m7.255s
  ```

  ```
  $ # Old workflow
  $ tox
  ...
  real    0m19.537s
  user    2m2.577s
  sys     0m7.972s
  ```
* `uv` manages installation of the appropriate versions of Python and all pip
  packages, greatly simplifying the  creation of PyRTL development
  environments
* This pins all developer tools to specific versions. Previously we only pinned
  documentation tools. This avoids confusing situations where lint errors
  suddenly appear just because a new version of `ruff` was released.

To run all tests with the new workflow:

```
$ uv run just tests
```

To build documentation with the new workflow:

```
$ uv run just docs
```

This replaces the old `tox` based workflow with a new `just` based workflow. `tox` is no longer needed since `uv` manages all virtual environments. `just` is used instead, as a simple portable task runner.


fdxmw:
This PR significantly changes the PyRTL developer workflow, but I think the advantages are worth it. Please let me know if you have questions or comments, otherwise I'll merge this in about a week.

PR #481, opened 2026-04-24
Floating point implementation refactoring

fdxmw:
This builds on the `wire_struct` improvements in #480.

* Define canonical `wire_struct` `Float` types, and use them as the source of
  truth for the number of exponent and mantissa bits. See `Float16` for an
  example. All operators are generic, and accept any of these `Float` types as
  input, and produce the same `Float` type as output. So if you add a `Float16`
  with a `Float16`, you get a `Float16`. If the operand types do not match, a
  `PyrtlError` is raised.

  This lets us remove `_fp_wire_struct`, `FPTypeProperties`,
  `FloatingPointType`, `PyrtlFloatConfig` because operands now carry their own
  metadata and describe themselves. For example, we can inspect the operand's
  mantissa bitwidth directly with `operand.mantissa.bitwidth`, rather than
  passing around another data structure.

* Remove the `FloatOperations` classes. They are no longer needed because the
  top-level `add`, `sub`, and `mul` functions are generic, and work with any
  matching `Float` types. The default rounding mode can be set with
  `set_default_rounding_mode`, and can be changed for individual operations.

* Rename the package from `pyrtlfloat` to just `float`. The full package name
  was `pyrtl.rtllib.pyrtlfloat`, which was redundant.

* Rename `mul` to `mult` for symmetry with `signed_mult`.

* Rename `make_inf` and friends to `make_inf_like`. These functions now take an
  operand that specifies the desired output bitwidths, similar to
  `np.ones_like`.

* Remove `make_zero`. It is clearer to directly set the exponent and mantissa
  to `0`.

* Use `dataclass` to simplify some class definitions.

* Use absolute imports (https://peps.python.org/pep-0008/#imports).

* Rename some variables to improve readability.

* Update `rtllib.rst` to generate Sphinx documentation for the `float` package.


fdxmw:
@gaborszita Here's the refactoring changes I had in mind for `pyrtlfloat`, please take a look when you get a chance and let me know what you think. The biggest change is that all operators now use `wire_structs` as their inputs and outputs, which lets us remove a bunch of data structures and code.

I also tried to clean up the documentation a bit, you can see what it currently looks like at https://pyrtl--481.org.readthedocs.build/en/481/rtllib.html#module-pyrtl.rtllib.float

gaborszita:
Than you for your suggestions, I'll take a look at them this week!

gaborszita:
I looked over the changes, they look good to me, thank you for cleaning this up! I agree, this is a much cleaner way to pass the properties of floating point numbers than always passing `fp_type_props`. We can merge it, or is there any other improvements I should do?

gaborszita:
The documentation also looks much better, thank you for fixing this too! The examples you added help a lot.

fdxmw:
> I looked over the changes, they look good to me, thank you for cleaning this up! I agree, this is a much cleaner way to pass the properties of floating point numbers than always passing `fp_type_props`. We can merge it, or is there any other improvements I should do?

Great, I'll merge this PR! You don't need to do anything, though you're always welcome to add new features or clean up the code some more :)

PR #482, opened 2026-05-30
Initial `StateRegister` implementation.

fdxmw:
A `StateRegister` is just like a `Register`, except its bitwidth is defined by an `IntEnum`, and `render_trace` displays the `IntEnum`'s name instead of its `int` value.

This also changes `enum_name` so it converts unknown `int` values to string with `hex`, instead of throwing a `ValueError`.


fdxmw:
@timsherwood Thinking about this more, maybe this feature should be added to `Register` instead of defining a new subclass? Currently, the subclass doesn't seem to do enough to justify its existence.

fdxmw:
> @timsherwood Thinking about this more, maybe this feature should be added to `Register` instead of defining a new subclass? Currently, the subclass doesn't seem to do enough to justify its existence.

Okay I pushed another commit that adds a `State` option to `Register`'s constructor, instead of defining a new subclass. I like this new approach much more. I'll squash and merge this next week if nobody has comments.

PR #484, opened 2026-06-05
Fix Wavedrom in Jupyter Notebook

ryoon:
* Use HTML tags instead of jQuery.
  Tested with Jupyter Notebook 7.5.4.

There may be better solution, anyway please fix the error without jQuery.

### DIFF
--- a/pyrtl/simulation.py
+++ b/pyrtl/simulation.py
@@ -1775,16 +1775,12 @@ class SimulationTrace:
             html_elem = HTML(htmlstring)
             display(html_elem)
             # print(htmlstring)
-            js_stuff = """
-            $.when(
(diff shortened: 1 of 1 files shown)


fdxmw:
Thank you for fixing this! I tested this out and it works on my computer. I also added a small change that seems to fix a longstanding issue where traces don't render the first time a `render_trace()` cell executes.

Please let me know if the additional change looks good to you, and then I'll merge this pull request.

Edits: I updated the fix so it periodically checks if WaveDrom is loaded before calling `ProcessAll()`. I now see traces immediately when I load a Jupyter notebook and select `Run > Run All Cells`. Also, I'd like you to take a look at this change, if you have time, because I don't know anything about JavaScript :)

ryoon:
Hi,

I am not a JavaScript professional sadly.
Your fix works fine for me.
Please merge this pull request.

Thank you.

PR #492, opened 2026-07-11
`name_scope` is a context manager that prepends prefixes to `WireVector` and  `MemBlock` names

fdxmw:
This helps avoid name collisions in larger designs.


fdxmw:
See [documentation](https://pyrtl--492.org.readthedocs.build/en/492/basic.html#naming) for some examples

PR #495, opened 2026-07-15
Apply `ConsecutiveSlice` optimizations to `CompiledSimulation` for ~46% reduction in compile time.

fdxmw:
For pyrtlnet, this reduces `CompiledSimulation`'s startup time from 2.96 seconds to 1.6 seconds. It does not affect simulation speed, so the compiler was optimizing the inefficient code we were generating before, but we were paying for those optimizations with startup time.

Also: Simplify the interface for `make_consecutive_slices` and move the `shift` helper in `simulation.py` to toplevel so we can call it from `compilesim.py`. Add some comments and emit bitmasks in hex so they're easier to identify.

### DIFF
--- a/pyrtl/compilesim.py
+++ b/pyrtl/compilesim.py
@@ -16,10 +16,10 @@ from pyrtl.helperfuncs import infer_val_and_bitwidth
 from pyrtl.memory import MemBlock, RomBlock
 from pyrtl.pyrtlexceptions import PyrtlError, PyrtlInternalError
 from pyrtl.simulation import (
-    ContiguousSlice,
+    ConsecutiveSlice,
     SimulationTrace,
     _trace_sort_key,
-    make_contiguous_slices,
+    make_consecutive_slices,
     shift,
 )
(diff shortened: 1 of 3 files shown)


fdxmw:
Example: the old code would generate:
```
((1&(w4134_layer1matmulreg10[0]>>0))<<0)|
 ((1&(w4134_layer1matmulreg10[0]>>1))<<1)|
 ((1&(w4134_layer1matmulreg10[0]>>2))<<2)|
 ((1&(w4134_layer1matmulreg10[0]>>3))<<3)|
 ((1&(w4134_layer1matmulreg10[0]>>4))<<4)|
 ((1&(w4134_layer1matmulreg10[0]>>5))<<5)|
 ((1&(w4134_layer1matmulreg10[0]>>6))<<6)|
 ((1&(w4134_layer1matmulreg10[0]>>7))<<7)
```
while the new code generates:

```
(w4134_layer1matmulreg10[0] & 0xFF)
```

PR #499, opened 2026-07-23
Fix a bug in `CompiledSimulation`'s `arm64` multiplication macro. Enable GitHub actions testing on `macos-latest` to catch these bugs in the future. Also:

fdxmw:
* Disable `doctests` for `output_to_svg` and `graphviz_detailed_namer` because they depend on `graphviz`, which may not be installed.

* Update `pytest` command line to print the names of skipped tests.

### DIFF
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -5,19 +5,30 @@ on:
   - pull_request
 
 jobs:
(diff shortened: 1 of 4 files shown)


fdxmw:
Python's load-bearing comma strikes again :(

PR #500, opened 2026-07-24
Enable Python test GitHub actions workflow on the `windows-latest` runner and fix several Windows compatibility issues:

fdxmw:
* Rename the `cp437` renderer constants to `utf-8-basic`. The old constants were actually Unicode, and not actually code page 437, so they weren't working as advertised. This removes code page 437 support, since nobody seems to be using it.

* Make `render_trace`'s platform detection code more reliable.

* Fix `test_examples.py`'s path handling so it works on Windows.

* Don't use `render_trace` in doctests outside of `Simulation`, and disable `Simulation` doctests when Unicode is not supported. `Simulation` doctest examples expect Unicode output for `render_trace`.

* Fix `examples` so they don't require Unicode.

* Fix `justfile` to work on Windows.

* Add doctests for `Simulation` and `Simulation.step`.


fdxmw:
I also tested this PR manually by running `uv run just presubmit` and some `examples` on Windows, in `cmd.exe`, PowerShell, and Cygwin.

PR #504, opened 2026-07-31
Improve `render_trace`'s ASCII output.

fdxmw:
This makes it easier to visually distinguish high and low, for 1-bit signals.


fdxmw:
The `Files changed` tab has a nice side-by-side screenshot comparison.

PR #508, opened 2026-08-01
Polish GitHub Pages styling

timsherwood:
Waveform screenshot backgrounds now extend to fill their containers for a seamless terminal look.

### DIFF
--- a/www/_static/custom.css
+++ b/www/_static/custom.css
@@ -4,6 +4,7 @@ body {
 
 div.body h1 {
     font-size: 120%;
+    margin-top: 3rem;
 }
 
 /* Spacing adjustments. */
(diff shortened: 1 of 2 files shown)


www/_static/custom.css
```diff
@@ -91,6 +91,14 @@ div.document {
     margin-bottom: 1.5rem;
 }
 
+/* Waveform screenshots: extend dark background across full width. */
+.image-reference {
+    display: block;
+    background-color: #000000;
```
fdxmw: This can be written more concisely as `#000`

fdxmw (commented):
You'll need to resolve conflicts before we can merge, looks like the first commit should be removed?

Issue #511, opened 2026-08-06
False-Positive Type Errors in Register


PigBrainOverflow:
Hi developers of Pyrtl!

Inside `conditional_assignment` context, `reg.next |= value` will cause
`Cannot assign to attribute "next" for class "Register" Argument of type "_Next" cannot be assigned to parameter "value" of type ...`

The code is valid at runtime and I use Pyright with strict type checking mode.

Thanks.

fdxmw:
Thanks for the report! I looked into this a bit, here's my current thoughts.

The specific issue you're pointing out seems relatively easy to fix, and I'd be happy to review a PR for that, but I'm not yet convinced that strictly typechecking PyRTL (or Python in general) is a good idea. PyRTL currently has over a thousand type errors, and it will take considerable effort to address them. Some of those errors are straightforward to fix, but many are non-trivial.

Consider [this adder example code](https://github.com/UCSBarchlab/PyRTL/blob/development/www/examples/example-adder.py#L21-L26):

```python
sum = [None] * a.bitwidth
for i in range(a.bitwidth):
    sum[i], cout = fa(a[i], b[i], cin)
    cin = cout

full_sum = pyrtl.concat_list(sum)
```

This initializes a `list` with `None`s, then replaces each `None` with a `WireVector`. Static typechecking does not approve:

```
error[invalid-argument-type]: Argument to function `concat_list` is incorrect
  --> www/examples/example-adder.py:26:34
   |
26 |     full_sum = pyrtl.concat_list(sum)
   |                                  ^^^ Expected `list[WireVector | int | str]`, found `list[None | Unknown]`
info: element `None` of union `None | Unknown` is not assignable to `WireVector | int | str`
```

(the typechecker thinks there could still be `None`s in the `list`). Cases like this are challenging because they're hard to make typesafe without harming readability. We shouldn't add annotations only for the typechecker's benefit.

PyRTL's type hints currently serve mostly as documentation, and that documentation is often intentionally incomplete: users shouldn't need to know about internal classes like `_MemIndexed` or `WrappedWireVector`, even though we return instances of these classes to users. This adds to the challenge, because I don't want to end up maintaining two type systems, one for documentation and another for type checking.

[...]

PR #513, opened 2026-08-22
Feature: more typing

PigBrainOverflow:
#511

- Fix the `Register.next` setter annotation to reflect that augmented assignment passes a `Register._Next` object back through the setter.
- Change `match_bitwidth`'s return type to `Iterator[WireVector]`, matching its current generator-based implementation.
- Change `concat_list`'s parameter type from `list[WireVectorLike]` to `Sequence[WireVectorLike]`.
- New implementation of [adder](www/examples/example-adder.py) and [fir](www/examples/example-fir.py).

The typing changes are annotation-only and do not change runtime behavior.

This PR focuses only on fixing the identified type annotations. I have not added repository-wide type-regression testing yet, since doing that robustly is a substantially larger piece of work.

A useful regression test would first require choosing a representative baseline and defining what it means for the typing state to become "no worse." This is not necessarily equivalent to comparing the number of diagnostics: a future change might eliminate several existing errors while introducing a smaller number of different errors, and it is not obvious whether that should be considered an improvement or a regression.

It would therefore be better to design the baseline and comparison policy separately, rather than introducing a fragile diagnostic-counting test as part of this PR.

### DIFF
--- a/www/examples/example-fir.py
+++ b/www/examples/example-fir.py
@@ -1,4 +1,5 @@
 import pyrtl
+import itertools
 
 # # Finite impulse filter example.
 
@@ -7,10 +8,11 @@ def fir(x: pyrtl.WireVector, bs: list[int]):
     rwidth = x.bitwidth  # Bitwidth of the registers.
     ntaps = len(bs)  # Number of coefficients.
 
+    # Create a chain of registers.
     regs = [pyrtl.Register(rwidth) for _ in range(ntaps - 1)]
-    for i, reg in enumerate(regs):
-        reg.next <<= x if i == 0 else regs[i - 1]
-
+    regs[0].next <<= x
+    for prev, curr in itertools.pairwise(regs):
+        curr.next <<= prev
     zs = [x, *regs]
 
     # Produce the final sum of products.
--- a/www/examples/example-fir.py
+++ b/www/examples/example-fir.py
@@ -7,9 +7,11 @@ def fir(x: pyrtl.WireVector, bs: list[int]):
     rwidth = x.bitwidth  # Bitwidth of the registers.
     ntaps = len(bs)  # Number of coefficients.
 
-    zs = [x] + [pyrtl.Register(rwidth) for _ in range(ntaps - 1)]
-    for i in range(1, ntaps):
-        zs[i].next <<= zs[i - 1]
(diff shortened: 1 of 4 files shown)


www/examples/example-fir.py
```diff
@@ -7,9 +7,11 @@ def fir(x: pyrtl.WireVector, bs: list[int]):
     rwidth = x.bitwidth  # Bitwidth of the registers.
     ntaps = len(bs)  # Number of coefficients.
 
-    zs = [x] + [pyrtl.Register(rwidth) for _ in range(ntaps - 1)]
-    for i in range(1, ntaps):
-        zs[i].next <<= zs[i - 1]
+    regs = [pyrtl.Register(rwidth) for _ in range(ntaps - 1)]
+    for i, reg in enumerate(regs):
+        reg.next <<= x if i == 0 else regs[i - 1]
```
fdxmw: Hmm, I find this harder to read than the original. What do you think about something like this?

```python
regs = [pyrtl.Register(rwidth) for _ in range(ntaps - 1)]
regs[0].next <<= x
for prev, curr in itertools.pairwise(regs):
    curr.next <<= prev
zs = [x, *regs]
```

fdxmw:
Thanks for putting this together!

> This PR focuses only on fixing the identified type annotations. I have not added repository-wide type-regression testing yet, since doing that robustly is a substantially larger piece of work.

I agree that regression testing should be added in a separate PR, and that defining exactly what "no worse" means is quite hard. Some thoughts:

1. Having some tests is generally better than having no tests. I'd rather have some simple regression tests soon, than having perfect regression testing in the indefinite future :) I think I'd start with a short allow-list of examples, tests, etc that currently pass typechecks, and add checks to be sure those stay type-safe in the future.
2. Type-regression testing on a type-unsafe codebase is a hard problem, but it's a meta-problem. I think our time and energy are better spent on the underlying problem of making PyRTL more type-safe, than finding the best solution to this meta-problem.

fdxmw (commented):
There are also some formatting errors, you should be able to reproduce them with `uv run just presubmit`

www/examples/example-fir.py
PigBrainOverflow: Cool! It's clearer. I will adopt it

